PR #4, opened 2020-04-07
Fixed easter egg hunt and onslaught. Fixed impact bomb

Dmitry450:


assets/src/ba_data/python/bastd/game/easteregghunt.py
```diff
@@ -147,7 +147,7 @@ def _on_egg_player_collide(self) -> None:
                 player = (spaz.getplayer()
                           if hasattr(spaz, 'getplayer') else None)
                 if player and egg:
-                    player.get_team().gamedata['score'] += 1
```
efroemling, line 150: Note: the fact that this error didn't get picked up by type-checking ('make mypy' or 'make check') tells me that the type-checker doesn't have types for these variables here. See https://github.com/efroemling/ballistica/wiki/Knowledge-Nuggets#beware-the-any-type
Don't worry about it for now, but in the future it could be good to first get the type checker detecting the issue and then fixing the issue (that way the type checker would warn us if other issues popped up in this code later).
Adding an `assert isinstance(player, ba.Player) or player is None` would inform the type checker what player is supposed to be. (which you could then verify with reveal_type())

assets/src/ba_data/python/bastd/game/easteregghunt.py
```diff
@@ -186,9 +186,13 @@ def _update(self) -> None:
...
         ypos = random.uniform(3.5, 3.5)
         zpos = random.uniform(-8.2, 3.7)
+        def _is_exists(egg: Egg) -> bool:
+            if egg.node is None:
+                return False
+            return egg.node.exists()
 
         # Prune dead eggs from our list.
-        self._eggs = [e for e in self._eggs if e]
+        self._eggs = [e for e in self._eggs if _is_exists(e)]
```
efroemling, line 195: Interesting; the original code is supposed to be doing the same thing as this; the .exists() method for the egg should return false when the node is None or an invalid reference, and `if actor` now should be the same as `if actor.exists()`. I'll take a look and see which of that is not happening...

efroemling:
Thanks a ton for the fixes! I'll try to get a 'Contributors' file set up soon and I'd love to have you add your name to that...

PR #11, opened 2020-04-08
Bots AI fixes, and some game modes editings

Dmitry450:
ba/_gameutils.py - added function:
```python3
def distance(Tuple[float, float, float], Tuple[float, float, float]) -> float:
```
In some game modes ```player.get_team()``` and ```flag.get_team()``` replaced by ```player.team``` and ```flag.team```

In some actors ```_handle_message_sanity_check``` moved into ```if __debug__:```

### DIFF
--- a/assets/src/ba_data/python/ba/_gameutils.py
+++ b/assets/src/ba_data/python/ba/_gameutils.py
@@ -22,12 +22,13 @@
 from __future__ import annotations
 
 from typing import TYPE_CHECKING
+import math
 
 import _ba
 from ba._enums import TimeType, TimeFormat, SpecialChar
 
 if TYPE_CHECKING:
-    from typing import Any, Dict, Sequence
+    from typing import Any, Dict, Sequence, Tuple
     import ba
 
 TROPHY_CHARS = {
@@ -46,6 +47,14 @@ def get_trophy_string(trophy_id: str) -> str:
         return _ba.charstr(TROPHY_CHARS[trophy_id])
     return '?'
 
+def distance(vec1: Tuple[float, float, float],
+             vec2: Tuple[float, float, float]) -> float:
+    """Find distance between two positions"""
+    xlen = vec1[0] - vec2[0]
+    ylen = vec1[1] - vec2[1]
+    zlen = vec1[2] - vec2[2]
+    xylen = math.sqrt(xlen**2 + ylen**2)
+    return math.sqrt(xylen**2 + zlen**2)
 
 def sharedobj(name: str) -> Any:
     """Return a predefined object for the current Activity, creating if needed.
--- a/assets/src/ba_data/python/bastd/actor/spazbot.py
+++ b/assets/src/ba_data/python/bastd/actor/spazbot.py
@@ -42,8 +42,7 @@ DEFAULT_BOT_HIGHLIGHT = (0.1, 0.3, 0.1)
 PRO_BOT_COLOR = (1.0, 0.2, 0.1)
 PRO_BOT_HIGHLIGHT = (0.6, 0.1, 0.05)
 
-def _my_vec3_dist(vec1: Tuple[float], vec2: Tuple[float]) -> float:
-    # FIXME - this solution works,
+def distance(vec1: Tuple[float], vec2: Tuple[float]) -> float:
     xlen = vec1[0] - vec2[0]
     ylen = vec1[1] - vec2[1]
     zlen = vec1[2] - vec2[2]
@@ -200,7 +199,7 @@ class SpazBot(basespaz.Spaz):
         closest = None
         assert self._player_pts is not None
         for plpt, plvel in self._player_pts:
-            dist = _my_vec3_dist((plpt.x, plpt.y, plpt.z), (botpt.x, botpt.y, botpt.z))
+            dist = distance((plpt.x, plpt.y, plpt.z), (botpt.x, botpt.y, botpt.z))
 
             # Ignore player-points that are significantly below the bot
             # (keeps bots from following players off cliffs).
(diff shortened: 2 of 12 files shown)


assets/src/ba_data/python/bastd/actor/spazbot.py
```diff
@@ -256,7 +257,9 @@ def update_ai(self) -> None:
                 target_pt_raw = ba.Vec3(*self.target_flag.node.position)
                 diff = (target_pt_raw - our_pos)
                 diff = ba.Vec3(diff[0], 0, diff[2])  # don't care about y
-                dist = diff.length()
+                dist = distance(
+                    (target_pt_raw.x, 0, target_pt_raw.z),
+                    (our_pos.x, 0, our_pos.z))
```
efroemling, line 262: Doesn't the old code do the same as this? (it gets the difference between target_pt_raw and our_pos, zeroes y, and then gets the length of that).

I'd like to use the ba.Vec3 class when possible since that is implemented in C++ so should be more efficient. If the math is broken in any way or it needs more functionality, I'm happy to fix it.

assets/src/ba_data/python/ba/_gameutils.py
```diff
@@ -46,6 +47,14 @@ def get_trophy_string(trophy_id: str) -> str:
...
     return '?'
 
+def distance(vec1: Tuple[float, float, float],
+             vec2: Tuple[float, float, float]) -> float:
+    """Find distance between two positions"""
+    xlen = vec1[0] - vec2[0]
+    ylen = vec1[1] - vec2[1]
+    zlen = vec1[2] - vec2[2]
+    xylen = math.sqrt(xlen**2 + ylen**2)
+    return math.sqrt(xylen**2 + zlen**2)
```
efroemling, line 57: Same note as down below; I'd prefer to go through ba.Vec3 as much as possible for functionality like this instead of doing raw number crunching in Python. Over time I will have more node attributes/etc. return ba.Vec3 instead of lists/tuples to hopefully make vector math as easy as possible. If there's anything coming out incorrect with the ba.Vec3 methods or anything that could be easier to use, please let me know.

assets/src/ba_data/python/bastd/actor/spazbot.py
```diff
@@ -324,7 +327,9 @@ def update_ai(self) -> None:
                      target_vel * dist_raw * 0.3 * self._lead_amount)
 
         diff = (target_pt - our_pos)
-        dist = diff.length()
+        dist = distance(
+            (target_pt.x, target_pt.y, target_pt.z),
+            (our_pos.x, our_pos.y, our_pos.z))
```
efroemling, line 332: Same here; old code should work now that ba.Vec3.length() is fixed. Please holler if not.

assets/src/ba_data/python/bastd/game/assault.py
```diff
@@ -176,7 +176,7 @@ def _handle_base_collide(self, team: ba.Team) -> None:
             return
 
         # If its another team's player, they scored.
-        player_team = player.get_team()
+        player_team = player.team
```
efroemling, line 179: Don't worry about it for now, but in the future could I ask that you fix the type-checking in cases like this in addition to the error? Mypy doesn't know what player_team is here, otherwise it would have raised an alarm for the incorrect get_team() call.

Basically, if you temporarily add a `reveal_type(player)` statement and then run `make mypy`, you'll probably see that player is considered `Any`. If you add an `assert isinstance(player, ba.Player)` right after player is assigned, this will tell the type checker that it is in fact a Player and it'll do better at catching errors (both this one and others in the future that may arise)

See info here: https://github.com/efroemling/ballistica/wiki/Knowledge-Nuggets#beware-the-any-type

assets/src/ba_data/python/bastd/game/capturetheflag.py
```diff
@@ -63,7 +63,8 @@ def reset_return_times(self) -> None:
         self.touch_return_time = float(
             self.activity.settings['Flag Touch Return Time'])
 
-    def get_team(self) -> ba.Team:
+    @property
+    def team(self) -> ba.Team:
```
efroemling, line 67: Nice! I am trying to migrate things like this to properties since those are more elegant looking in code..

Dliwk:
@efroemling One question. Why are you ignoring the y coordinate when calculating distance? If some modders create a multi-level map (e.g. BombDash Esher's Waterwall), the bots will be.. strange if a player is under or  above them.

efroemling:
The bot ai basically thinks in 2d; it is always trying to reach/attack/etc. a point on the same plane as itself. So if an enemy were standing above it, the bot would try to punch or bomb the point right below since we take y out of the equation. If we leave y in the equation, the bot will calculate larger distances and thus think you are farther from it and will run through that point, throw farther, etc. Neither behavior is really correct, but the first seems more predictable/debuggable to me. Otherwise getting blown up in the air/etc. might trigger strange behavior from bots.

You’ll also notice that any coop levels with bots are mostly flat for this reason :)

Hope that helps?…
-Eric

Dliwk:
Haha, you're right. I did not think that the bot would just run faster. But maybe it's worth checking the y-distance separately, not in equation? If it is too large (higher than bot can drop a bomb), the bot will ignore this player. This will help if the bot and player are not at neighboring levels. 
And I think this helps if bot wants to pickup flag.

efroemling:
These all sound like possibly good ideas, but I would rather hold off on logic changes for now. I think at some point it could be good to revisit the bot code (or spin off a separate one) and make it smarter about handling 3d terrain in general.

efroemling:
I'm going to go ahead and pull this in and just revert some of the stuff such as the distance() function since Vec3 should be working now...

assets/src/ba_data/python/bastd/game/assault.py
Dmitry450: Thanks I got it. I will try to use this in future.

PR #12, opened 2020-04-09
Fixed coop version of race

Dmitry450:


efroemling:
Looks good. Thanks!
Only one minor thing I might tweak; generally local variables don't have leading underscores so I might rename those to cur_time and start_time. (at first glance it made them look like instance variables)

PR #13, opened 2020-04-09
Fixed meteor shower. Fixed text in co-op score screen. Fixed TNT respawn

Dmitry450:


assets/src/ba_data/python/bastd/activity/coopscorescreen.py
```diff
@@ -1193,22 +1193,22 @@ def _show_world_rank(self, offs_x: float) -> None:
...
+            self._tournament_time_remaining_text = Text(
+                '',
+                position=(-360, -110 - 100),
+                color=(1, 1, 1, 0.7),
+                h_align=Text.VAlign.CENTER,
+                v_align=Text.VAlign.CENTER,
+                transition=Text.Transition.FADE_IN,
+                scale=1.6,
+                maxwidth=150,
+                transition_delay=2.0)
```
efroemling, line 1211: This is an interesting case where type-checking wasn't running for these statements.
if turns out that the `if self._tournament_time_remaining is not None` line threw things off.
If I added a reveal_type() call above that statement, Mypy would give me a value, but adding one under the `if` statement gave nothing. This shows that Mypy wasn't evaluating that code.

It turns out that Mypy considered self._tournament_time_remaining to be None and so considered that code to be unreachable and thus didn't type check it. This is because self._tournament_type_remaining was declared simply as None. If I give it a proper Optional[float] type, Mypy does evaluate the code under the `if` statement and properly flags the error.

assets/src/ba_data/python/ba/_gameactivity.py
```diff
@@ -1111,12 +1111,8 @@ def setup_standard_powerup_drops(self, enable_tnt: bool = True) -> None:
...
         self._standard_drop_powerups()
         if enable_tnt:
-            self._tnt_objs = {}
-            self._tnt_drop_timer = _ba.Timer(5.5,
-                                             _general.WeakCall(
-                                                 self._standard_drop_tnt),
-                                             repeat=True)
-            self._standard_drop_tnt()
+            self._tnt_spawners = {}
+            self._setup_standard_tnt_drops()
```
efroemling, line 1115: Nice cleanup!

assets/src/ba_data/python/bastd/game/meteorshower.py
```diff
@@ -164,7 +164,7 @@ def handlemessage(self, msg: Any) -> Any:
             # Augment standard behavior.
             super().handlemessage(msg)
 
-            death_time = ba.time()
+            death_time = ba.time(timeformat=ba.TimeFormat.MILLISECONDS)
```
efroemling, line 167: Thanks for fixing any breakage here, but I may go through and try to convert this to seconds. I'd love to standardize on using float seconds whenever possible just to avoid ugly ba.TimeFormat.MILLISECONDS args everywhere.

assets/src/ba_data/python/bastd/activity/coopscorescreen.py
```diff
@@ -1193,22 +1193,22 @@ def _show_world_rank(self, offs_x: float) -> None:
...
-                                                        v_align='center',
-                                                        transition='fade_in',
-                                                        scale=1.6,
-                                                        maxwidth=150,
-                                                        transition_delay=2.0)
+            self._tournament_time_remaining_text = Text(
+                '',
+                position=(-360, -110 - 100),
+                color=(1, 1, 1, 0.7),
+                h_align=Text.VAlign.CENTER,
```
efroemling, line 1206: A good little example of the benefit of trying to get type-checking functioning before applying a fix: h_align was being given a Text.VAlign value here. (Though this didn't cause any problems in this particular case.)

PR #14, opened 2020-04-09
Fix score sorting in co-op scorescreen activity

Dliwk:
If some scores are equal, I get TypeError while Python compares dict objects.
I think that it’s also necessary to sort by time, but it will take a lot of effort and I don’t know if it is necessary.

### DIFF
--- a/assets/src/ba_data/python/ba/_gameresults.py
+++ b/assets/src/ba_data/python/ba/_gameresults.py
@@ -189,7 +189,7 @@ class TeamGameResults:
             sval.append(team)
         results: List[Tuple[Optional[int],
                             List[ba.Team]]] = list(winners.items())
-        results.sort(reverse=not self._lower_is_better)
+        results.sort(reverse=not self._lower_is_better, key=lambda x: x[0])
 
         # Also group the 'None' scores.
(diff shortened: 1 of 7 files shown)


efroemling:
Interesting; I'd like to take a look at this since it may indicate that I broke something elsewhere. Do you have an easy way to reproduce this?

Dliwk:
I think I don't understand. Do you want to search for other places without specifying a sort key?

efroemling:
I am just curious to know more about why the error occurs without a sort key.  I don't think it used to in the 1.4 code (unless that was a bug that went unnoticed this whole time). So I was curious how I could reproduce the bug. Do I just need to play any co-op game twice and get the same score?

Dliwk:
Well, I found this bug just like that, I had the same results.

I think you know that Python compares the lists as follows: it compares first elements and if they are equal (which we did), then it starts to compare the second and so on (and the second element is a dict). If you specify a sort key, it will sort only by this key, and will not do anything if they are equal.

Dliwk:
I found four similar situations (sorting nested lists without specifying a key with elements that Python cannot compare). Three in `bastd` module, one in `ba`.

PR #15, opened 2020-04-09
Fix buy method in StoreBrowserWindow

Dliwk:
Confirm window did't appear.

Now it should work but.. I think ba.internal.serverget (i.e. ServerCallThread) doesn't work, it doesn't run callback.

### DIFF
--- a/assets/src/ba_data/python/bastd/ui/store/browser.py
+++ b/assets/src/ba_data/python/bastd/ui/store/browser.py
@@ -88,7 +88,7 @@ class StoreBrowserWindow(ba.Window):
 
         self._request: Any = None
         self._r = 'store'
-        self._last_buy_time = None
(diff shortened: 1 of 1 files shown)


Dliwk:
Hmm I can fix it... but I think what removing checking current activity in _run_callback method isn't the best way.
```python
activity = None if self._activity is None else self._activity()
if activity is None or activity.is_expired():
    return
```

efroemling:
> Hmm I can fix it... but I think what removing checking current activity in _run_callback method isn't the best way.
> 
> ```python
> activity = None if self._activity is None else self._activity()
> if activity is None or activity.is_expired():
>     return
> ```

You were right in that the code in _run_callback was messed up. I think I fixed it for type-checking and accidentally broke the logic. I just pushed a commit with a fix.

PR #16, opened 2020-04-10
Fix pause menu

Dliwk:
Functionality for adding custom menu entities didn't work correctly (in co-op sessions was no retry button).
I really don't know what this line do here:neutral_face:


efroemling:
> Functionality for adding custom menu entities didn't work correctly (in co-op sessions was no retry button).
> I really don't know what this line do here😐

I'm really not sure why that line was there either 🤔
It might have just been a complete accident while I was moving code around. 🤷🏼‍♂️
Thanks for the fix!

PR #18, opened 2020-04-11
Fix FFA session

Dliwk:
Hmm.... In these lines, the variable team_names is set to not None.
```python
if team_names is None:
    team_names = ['Good Guys']
if team_colors is None:
    team_colors = [(0.6, 0.2, 1.0)]
````
but in following line `self._use_teams` always True
```python
self._use_teams = (team_names is not None)
```


efroemling:
Thanks; this logic might have been left over from a while back when Session and Activity were a single class.. this stuff could definitely use a bit more cleaning up so maybe now is a good time to do that :)

PR #28, opened 2020-04-23
Small docs addition: "How to play custom music"

Dliwk:
# :scroll: Docs

Just a bit addition to docs about ba module. I figured out how to make custom music for about 15 minutes :sweat_smile:

### DIFF
--- a/assets/src/ba_data/python/ba/_music.py
+++ b/assets/src/ba_data/python/ba/_music.py
@@ -606,6 +606,8 @@ def setmusic(musictype: Optional[MusicType], continuous: bool = False) -> None:
     and also supports custom user soundtracks on specific platforms so the
     user can override particular game music with their own.
 
+    Pass ba.MusicType('<music_name>') to use custom music.
+
     Pass None to stop music.
 
     if 'continuous' is True the musictype passed is the same as what is already


Dliwk:
Oops ahah.. It doesn't work :disappointed: But how to play custom music?

efroemling:
Currently that function accepts a [MusicType](https://github.com/efroemling/ballistica/blob/master/docs/ba_module.md#class_ba_MusicType) enum which tells the game which a 'type' of music to play. The game has a default file it plays for each type and also the user can override that by creating their own 'Soundtrack'.

Custom soundtracks are implemented on Mac and Android currently.

Perhaps we should modify that function allow passing a filename in directly for cases where modders want to play a specific thing...

Dliwk:
Thanks; I know that this is an enum, but for some reason I thought that if I pass an argument to the constructor, it will create an "unnamed" type of music and this file will be played.

efroemling:
Sadly no; passing a value to an Enum's constructor will simply give you the Enum corresponding to that value. (so MusicType('menu') will give you MusicType.MENU, etc.)

efroemling:
Ok I just pushed a commit which cleans up ba._music stuff a bit so hopefully its easier to see what is going on.

Basically, to play specific music of your own you just need to not use the soundtrack system. So call ba.setmusic(None) and then create a 'sound' node to play whatever you want. See ba._music.MusicController._play_internal_music() for an example.

efroemling:
I'll go ahead and close this, but please don't hesitate to ask if you have any other questions about music playback or if I can clarify anything in the docs/etc.

PR #34, opened 2020-05-07
Added os.chdir() in ballistica_server.py

Awesome-Logic:
I realized the script couldn't be executed from other directories so I added the chdir function.
Now the script can be properly executed from anywhere.
If that was intended, feel free to reject the request. Also first pull request :)

### DIFF
--- a/assets/src/server/ballisticacore_server.py
+++ b/assets/src/server/ballisticacore_server.py
@@ -22,16 +22,21 @@
 """BallisticaCore server manager."""
 from __future__ import annotations
 
-import sys
-import os
 import json
+import os
 import subprocess
+import sys
 import time
-from threading import Thread, Lock, current_thread
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


Awesome-Logic:
The unix check seems to be failing due to some formatting error. I tried to format the files using make format but it's giving me an error related to 0 workers.

Awesome-Logic:
I had to manually format the files. `make format` seems broken for me

efroemling:
Sounds good to me. If someone ever has a strange need to use a server script with a different dist dir or something we can always add an explicit flag for that. Thanks!

efroemling:
> The unix check seems to be failing due to some formatting error. I tried to format the files using make format but it's giving me an error related to 0 workers.

Can you paste the exact error you are seeing here? (or traceback or whatever it gives you). I'd love to fix that for you. `make format` is such a time saver...

PR #35, opened 2020-05-07
Character-Character-Attack protection

Dliwk:
If chooser changes character too quickly, server just breaks down with some error (but it not restarts, just does not let anyone in).

This protection kicks players what changes characters too quickly and too much.

### DIFF
--- a/assets/src/ba_data/python/ba/_lobby.py
+++ b/assets/src/ba_data/python/ba/_lobby.py
@@ -34,7 +34,6 @@ if TYPE_CHECKING:
     from typing import Optional, List, Dict, Any, Sequence, Union
     import ba
 
-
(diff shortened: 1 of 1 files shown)


efroemling:
I would love to fix the underlying error if possible. Otherwise this same sort of attack may work elsewhere. Do you have the error message that is generated when the chooser switches too quickly?

Dliwk:
Most offen this is `Error: ThreadMessage list > 1000 for thread: unknown thread (sys-name="bs_headless")`, but sometimes console may be clean or contain something else.

efroemling:
Thanks for the info and for the PR.
I'll accept the PR for now, though I might remove it later if I can add protection in the underlying engine code.

PR #39, opened 2020-05-20
Minigame fixes

Dliwk:
1. In race, we have only one or two players in scoreboard (only that finished).
2. In Elimination, player's icon stays alive if he leaves.

Do we have any reason not to move base DieMessage handling to ba.Actor?

### DIFF
--- a/assets/src/ba_data/python/bastd/game/elimination.py
+++ b/assets/src/ba_data/python/bastd/game/elimination.py
@@ -166,7 +166,7 @@ class Icon(ba.Actor):
         if isinstance(msg, ba.DieMessage):
             self.node.delete()
         else:
-            return super().handlemessage(msg)
+            super().handlemessage(msg)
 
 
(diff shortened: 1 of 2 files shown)


efroemling:
Thanks for the heads-up on these bugs; I manually applied both fixes (though not through this pull request since some of the code had changed a bit).

In response to your question about DieMessage:

In the old code, ba.Actor would look for a 'node' attribute on itself, and if one was present it would handle DieMessage by deleting that node. This is the reason that Elimination icons were broken; they were expecting that code to still be there.

I wanted to change ba.Actor to be more general so there are no assumptions about what nodes it contains or what they are named.

There is still the ba.NodeActor type which has the old behavior; it wraps a ba.Node and will kill that node when it receives a DieMessage.  Another fix for the Elimination bug could have been to change the icons to inherit from that. (though I think just adding the DieMessage handler was simpler).

Thanks again!

PR #40, opened 2020-05-20
Fixed profile colors in ffa session and a bit cleanup

Dliwk:
Just fixed profile colors in FFA session (characters were white).


efroemling:
Thanks for the fix and the cleanup.
I'm trying to clean out any empty 'except Exception' blocks, so thanks for helping in that. (I put those in a long time ago before I knew it was a bad idea).
The guideline I follow these days is that if something catches 'Exception' it should always print or log the error so it won't go unnoticed. Silently-ignored errors should always be more specific exception types.

PR #42, opened 2020-05-21
Exceptions cleanup and fix FFA spawn

Dliwk:
Exceptions cleanup (like `raise Exception("emm is it player?")` -> `raise TypeError("emm is it player?")` and `except Exception` -> ...).

FFA player spawn when game begins ~has been fixed~ still works bad (all were spawn only at one-two points). I have no access to internal ba.Player methods, so I can't say more than that something wrong with ba.Player.node (i.e. ba.Player._nodeactor); in short, player._nodeactor return True if player has just joined but hasn't spawned yet (just at the moment of calculating the better position for its spawn).

P.S. **"playerID"** argument for ba.newnode('player') still in camelCase.

### DIFF
--- a/assets/src/ba_data/python/ba/_account.py
+++ b/assets/src/ba_data/python/ba/_account.py
@@ -83,7 +83,7 @@ def get_league_rank_points(data: Optional[Dict[str, Any]],
         assert isinstance(trophies_total, int)
         return trophies_total
     if subset is not None:
-        raise Exception('invalid subset value: ' + str(subset))
+        raise ValueError('invalid subset value: ' + str(subset))
 
     if data['p']:
         pro_mult = 1.0 + float(
--- a/assets/src/ba_data/python/ba/_achievement.py
+++ b/assets/src/ba_data/python/ba/_achievement.py
@@ -140,7 +140,7 @@ def get_achievement(name: str) -> Achievement:
     achs = [a for a in _ba.app.achievements if a.name == name]
     assert len(achs) < 2
     if not achs:
-        raise Exception("Invalid achievement name: '" + name + "'")
+        raise ValueError("Invalid achievement name: '" + name + "'")
     return achs[0]
 
 
@@ -396,7 +396,7 @@ class Achievement:
             v_attach = Text.VAttach.TOP
             attach = Image.Attach.TOP_CENTER
         else:
-            raise Exception('invalid style "' + style + '"')
+            raise ValueError('invalid style "' + style + '"')
 
         # Attempt to determine what campaign we're in
         # (so we know whether to show "hard mode only").
--- a/assets/src/ba_data/python/ba/_activity.py
+++ b/assets/src/ba_data/python/ba/_activity.py
@@ -166,14 +166,14 @@ class Activity(DependencyComponent, Generic[PlayerType, TeamType]):
 
         session = _ba.getsession()
         if session is None:
-            raise Exception('No current session')
+            raise RuntimeError('No current session')
         self._session = weakref.ref(session)
 
         # Preloaded data for actors, maps, etc; indexed by type.
         self.preloads: Dict[Type, Any] = {}
 
         if not isinstance(settings, dict):
-            raise Exception('expected dict for settings')
+            raise TypeError('expected dict for settings')
         if _ba.getactivity(doraise=False) is not self:
             raise Exception('invalid context state')
 
@@ -292,8 +292,8 @@ class Activity(DependencyComponent, Generic[PlayerType, TeamType]):
         (internal)
         """
         if self.has_begun():
-            raise Exception('This should only be called for Activities'
-                            'that have not yet begun.')
+            raise RuntimeError('This should only be called for Activities'
+                               'that have not yet begun.')
         if not self._should_end_immediately or force:
             self._should_end_immediately = True
             self._should_end_immediately_results = results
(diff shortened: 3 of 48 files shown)


efroemling:
Looks good; thanks!

I'll make a note to take a look at the FFA spawn point issue. I don't think any of that logic should be depending on internal functionality. It should just be looking at the available spawn points and picking a few random ones and going with whatever is farthest from others. Perhaps something in the math broke when I switched it from raw tuples to ba.Vec3 values.

Dliwk:
No-no-no, I already checked ba.Map.get_ffa_start_position, its logic works fine (for example, if replace player.node.position to player.actor.node.position while getting player positions all will works, but I think it isn't best way).

efroemling:
Oh; interesting. I think I might see the problem. I wanted to avoid using player.actor.node since spawn-point-selection logic should not care about the type or internals of the player's actor. (Currently they are all PlayerSpaz actors but they might not always be).
Each player has a 'player' node with a position attribute. Whatever actor the player is assigned should be getting wired up to drive that attribute. In the case of a PlayerSpaz, it looks like it is the node's 'torso_position'. (bastd/actor/playerspaz.py line 97)
So perhaps node.torso_position is not returning a correct value at the start of the round?.. (while node.position is?)
I'll make myself a note to look into this, but it might be worth opening an Issue to track if if you'd like.

efroemling:
Thinking about this more: this could also be an order-of-operations issue.  The connectattr() function sets a target attribute based on the driver attribute's current value and then subsequently updates the target each sim step. However, if the source value is explicitly changed after the connectattr() call, the target won't be updated immediately. GameActivity.spawn_player_spaz() currently creates the PlayerSpaz actor and *then* moves it into place by sending it a StandMessage, so player.node.position is probably still set to the original spawn location at that point 🤔
We might just need to update PlayerSpaz to explicitly update the player node position after it handles a StandEvent() (probably by just calling the connectattr() call again)

Dliwk:
Why? player.node.position shouldn't return any correct position if player have not spawned yet. We may just ignore all non-spawned players when calculating its spawn position (but we can't detect non-spawned/non-existing players because player.node.exists() return True at this moment). Player shouldn't count itself to get spawn position...

efroemling:
I think only living players are being looked at here (or if not, then that should be the case).
The problem is that player.node.position is invalid even *after* they have spawned.. and until the next sim step runs.
Basically what happens is:
For each player at the start of the round:
- PlayerSpaz is created with a default location (probably 0,0,0 or close to it)
- The PlayerSpaz calls connectattr() to drive player.node.position. This sets its value to the current one (close to 0,0,0) and wires it to be updated each step.
- The PlayerSpaz then looks at existing players' node.position to see where it should be placed at. Those values are all still close to 0,0,0.
- The PlayerSpaz is sent a StandMessage to move to the chosen spawn location. This does not, however, update the player.node.position immediately

So at spawn time, I imagine each PlayerSpaz is calculating a spawn point thinking that all other existing players are close to 0,0,0.  Hopefully that makes sense?..

Dliwk:
Oh, I understood. It is really likely that the player believes that all is at the same point. 
But... it worked in 1.4 :thinking:

efroemling:
Yes, in 1.4 the code was accessing actor.node.position directly. I think that is why it worked.
I just wanted to abstract the code to not assume any particular type of actor.
I think making PlayerSpaz handle the StandMessage better will fix this (having it update the player.node.position immediately so they stay in sync). I can give that a try in a few minutes...

efroemling:
I just added a print statement in get_ffa_start_position() where it is looking through player positions, and it appears my assumption is correct (that player.node.position is an old value):
```
player.node.position is (0.0, 1.0999999046325684, 0.0) and actor.node.position is (-5.0546875, 5.278906345367432, -8.3515625)
player.node.position is (0.0, 1.0999999046325684, 0.0) and actor.node.position is (7.453125, 5.278906345367432, -7.94140625)
player.node.position is (0.0, 1.0999999046325684, 0.0) and actor.node.position is (-5.0546875, 5.278906345367432, -8.3515625)
player.node.position is (0.0, 1.0999999046325684, 0.0) and actor.node.position is (7.453125, 5.278906345367432, -7.94140625)
player.node.position is (0.0, 1.0999999046325684, 0.0) and actor.node.position is (6.65625, 5.306250095367432, -9.140625)
player.node.position is (0.0, 1.0999999046325684, 0.0) and actor.node.position is (-5.0546875, 5.278906345367432, -8.3515625)
player.node.position is (0.0, 1.0999999046325684, 0.0) and actor.node.position is (7.453125, 5.278906345367432, -7.94140625)
player.node.position is (0.0, 1.0999999046325684, 0.0) and actor.node.position is (6.65625, 5.306250095367432, -9.140625)
player.node.position is (0.0, 1.0999999046325684, 0.0) and actor.node.position is (7.21484375, 5.278906345367432, -8.6328125)
player.node.position is (0.0, 1.0999999046325684, 0.0) and actor.node.position is (-5.0546875, 5.278906345367432, -8.3515625)
player.node.position is (0.0, 1.0999999046325684, 0.0) and actor.node.position is (7.453125, 5.278906345367432, -7.94140625)
player.node.position is (0.0, 1.0999999046325684, 0.0) and actor.node.position is (6.65625, 5.306250095367432, -9.140625)
player.node.position is (0.0, 1.0999999046325684, 0.0) and actor.node.position is (7.21484375, 5.278906345367432, -8.6328125)
player.node.position is (0.0, 1.0999999046325684, 0.0) and actor.node.position is (5.75, 5.306250095367432, -8.875)
```
I'll write a fix for this and get a commit in; it shouldn't take long.

efroemling:
Ok the fix for this is in d6ccf2b12dfdd12303afecac165f9ca8de245c81

PR #61, opened 2020-06-17
Update CHANGELOG.md for better look in Markdown

Dliwk:
- Replaced ballistica.net with  [ballistica.net](https://ballistica/net)
- Replaced \_ with \\_; some text was _italic_ because of \_ in object names like \_disconnectClient.


efroemling:
Thanks!

PR #91, opened 2020-07-04
Improving bomb.py

Benefit-Zebra:
## Description
Making improvement to bomb.py in order to prevent bombs from getting hittype **punch**. 

## Type of Changes
|   | Type |
| ✓  | :bug: Bug fix  |
| ✓  | :hammer: Refactoring  |


Benefit-Zebra:
I have also changed the name of variable **ispunch** to **ispunched** since it sounds more appropriate like : _if bomb not is punched_, than _if bomb not is punch_.

assets/src/ba_data/python/bastd/actor/bomb.py
```diff
@@ -1005,12 +1005,16 @@ def arm(self) -> None:
...
 
     def _handle_hit(self, msg: ba.HitMessage) -> None:
-        ispunch = (msg.srcnode and msg.srcnode.getnodetype() == 'spaz')
+        ispunched = (msg.srcnode and msg.srcnode.getnodetype() == 'spaz')
 
         # Normal bombs are triggered by non-punch impacts;
         # impact-bombs by all impacts.
-        if (not self._exploded and not ispunch
+        if not ispunched and (not self._exploded
                 or self.bomb_type in ['impact', 'land_mine']):
```
efroemling, line 1013: Ok, I'm thinking about this and I think neither of these are right.

In my old code, a bomb that has already exploded but gets punched and is a mine/impact bomb will incorrectly explode a second time:
```python
>>> exploded = True
>>> ispunched = True
>>> is_mine_or_impact = True
>>> bool(exploded and not ispunched or is_mine_or_impact)  # My old logic
True  # Oops; exploding even though we exploded already.
```

In your code, a bomb that has already exploded but gets hit with a non-punch and is a mine/impact bomb will incorrectly explode a second time:
```python
>>> exploded = True
>>> ispunched = False
>>> is_mine_or_impact = True
>>> bool(not ispunched and (not exploded or is_mine_or_impact))  # Your logic
True  # Oops; exploding even though we exploded already
```

I believe the correct logic (and what I think I intended originally even though I messed it up) is:
```python
if (not self._exploded and (not ispunched or self.bomb_type in ['impact', 'land_mine'])):
```
This way we will never explode once we've already exploded and for bombs that aren't impact/land-mines we will ignore punches.

It seems like the underlying problem with the land-mine situation might be the logic below here that causes it to inherit the hit_type of what is setting it off. I don't remember exactly why that logic exists but I think I will take it out unless I can remember a strong reason to keep it. It doesn't make sense that a land-mine's explosion should be treated as a punch just because the land-mine was set off by a punch, etc. 🤔

I'll go ahead and merge this and then make those adjustments.. Thanks again for the contribution.

efroemling:
> I have also changed the name of variable **ispunch** to **ispunched** since it sounds more appropriate like : _if bomb not is punched_, than _if bomb not is punch_.

Works for me.  I think in my mind I had considered 'ispunch' to be referring to the hit message and not the bomb, but it works both ways. 👍🏻

PR #99, opened 2020-07-09
Update _ba.py module

Benefit-Zebra:
## Description
Added extra_acceleration as a ba.Node attribute in _ba.py since it was missing.

## Type of Changes
|   | Type |
| ✓  | :hammer: Refactoring  |


Benefit-Zebra:
I have sorted the ba.Node attributes according to categories so that they are easy to find. I was unable to categorize few, so I putted them in Miscellaneous.

efroemling:
Sorry to be holding this up, but this file is auto-generated and so I'm hoping to get the script added to the repo which generates this file so we can properly make the changes there as part of this PR...

Benefit-Zebra:
Ok @efroemling, till then, I will keep this PR as it is.

PR #115, opened 2020-07-16
Use -S flag for /usr/bin/env

alibo:
## Steps

- [X] Add yourself to CONTRIBUTORS if you are a new contributor.
- [X] Add a CHANGELOG entry describing what your PR does. (I've added it to `Unreleased` section based on https://keepachangelog.com/en/1.0.0/) 
- [X] Ensure `make preflight` completes successfully. (`PREFLIGHT SUCCESSFUL!`)
- [X] Write a good description on what the PR does.

## Description

If you run `./bombsquad_server` (instead of `python3.7 <file>`), it returns the below error because of using `-O` flag for `python3.7` in the shebang line:

```
$ ./bombsquad_server
/usr/bin/env: ‘python3.7 -O’: No such file or directory
/usr/bin/env: use -[v]S to pass options in shebang lines
```

This PR adds `-S` flag to `/usr/bin/env` to resolve this issue.

> **-S, --split-string=S**
>               process and split S into separate arguments; used to pass
>               multiple arguments on shebang lines
> https://man7.org/linux/man-pages/man1/env.1.html

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
|   | :sparkles: New feature |
|   | :hammer: Refactoring  |
|   | :scroll: Docs |

## Related Issue

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### Unreleased
+- Fixed the shebang line in `bombsquad_server` file by using `-S` flag for `/usr/bin/env`.
+
 ### 1.5.22 (20139)
 - Button and key names now display correctly again on Android (and are cleaned up on other platforms too).
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -14,3 +14,6 @@
 ### Benefit-Zebra
 - Unofficial BombSquad Bug Finder
 - Code Cleanup
+
+### Ali Borhani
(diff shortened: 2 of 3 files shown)


efroemling:
Good catch, and thanks!
Apparently some unix flavors such as macOS read the shebang line in as multiple arguments while others such as linux read it as a single arg. I was testing on macOS which is why this slipped by me. It looks like the -S option works for both though 👍

PR #127, opened 2020-07-25
Fixing Curse powerup box corner case

Benefit-Zebra:
## Description
When we play any game such as Last Stand, Football, etc. If a _curse box happens to be at last_, **after 8 seconds**, health powerup spawns. This usually irritates in games such as Last Stand. 

Usually maps have 4 powerup spawn postions, but Rampage map has 2.

I have updated the script in which if a **Curse box**, _happens to be at odd postion(1st or 3rd position)_, it will definitely spawn. If it _happens to be at even position_, there's a _50 percent chance of spawning_ it depending on User's Luck. 

**Feel free to reject this.** 

## Type of Changes
| ✓  | :bug: Bug fix  |


Benefit-Zebra:
I don't know, if **Pylint and Mypy hates me**. I can't do preflight or pylint and mypy checks on my phone.

TornadoesPride:
Ever tried like changing the curse box count to ZERO? Cuz if you didn't do try that and it's much helpful😐

Benefit-Zebra:
If I change Curse box count to Zero, it will not at all spawn. I want that it will be depended on User's Luck if it has to spawn on Even Position.

PR #130, opened 2020-07-25
Impact bomb owner fix

### LINKED ISSUE #129: No Impact blast when Hit owner
When we throw a impact bomb, and it touches us(owner), it does not explode. It does cause damage when the spaz is inside the range of Impact bomb when it touches ground and explodes. Here is the [video](https://drive.google.com/file/d/1-MYN5PSgP7wRi747w7KD36Hi7_T4ZDOb/view?usp=drivesdk) for better understanding.

Benefit-Zebra:
## Description
Fixed the bug which caused impact bomb to not blast on its owner.  I have commented out those lines. 

## Type of Changes
|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue

Closes #129

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### Unreleased
+- Fixed a impact bomb bug which caused impact bomb to not blast on its owner.
+
 ### 1.5.23 (20146)
 - Fixed the shebang line in `bombsquad_server` file by using `-S` flag for `/usr/bin/env`.
 - Fixed a bug with hardware keyboards emitting extra characters in the in-game console (~ or F2)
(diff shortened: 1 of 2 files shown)


efroemling:
Thanks for the submit, but I don't think this is a good idea for now. Explained [here](https://github.com/efroemling/ballistica/issues/129#issuecomment-663916936).
Perhaps in 1.8 we can address this in a way that doesn't break anything.

Also, for future reference, if you can resolve CI issues it makes it easier for me to accept PRs. The failure here simply mentions that 'node_delegate' is unused after your changes, so removing or commenting out that variable would cause the ci check to succeed.

Issue #137, opened 2020-07-29
Add a other languages for internal keyboard please


KoLenka12:
For example: Arabic, Russian, Spanish, etc.

Benefit-Zebra:
Adding other languages to Internal Game Keyboard will not be ideal since these languages do have extra symbols which will cause the layout go out of space. So the layout will have to be reformatted for each language. Also not all languages are able to be displayed on all devices.

But the decision of doing it depends on @efroemling.

Dliwk:
Hmm, maybe not bad idea might be some API for adding internal keyboard language? And who want may just use mod what support his language. For example:

```python
# ba_meta export keyboard
class RussianKeyboard(ba.Keyboard):
    pages = [[
        ('й', 'ц', 'у', 'к', 'е', 'н', 'г', 'ш', 'щ', 'з', 'х', 'ъ'),
        ('ф', 'ы', 'в', 'а', 'п', 'р', 'о', 'л', 'д', 'ж', 'э'),
        ('я', 'ч', 'с', 'м', 'и', 'т', 'ь', 'б', 'ю')
    ]]
```

It should support many languages...

Dliwk:
Closing now; simple [keymap-creating API][wiki] is available, so it's easy to create your own.

[wiki]: https://github.com/efroemling/ballistica/wiki/Knowledge-Nuggets#hello-world-3-creating-a-new-keyboard

PR #123, opened 2020-07-20
BombSquad Exclusive Emojis

Benefit-Zebra:
## Description
Internal Game Keyboard now has a new Shiny Emoji widget.
Also added Continuous CAPITAL letters feature. 

## Type of Changes
|   | Type |
| ------------- | ------------- |
| ✓  | :sparkles: New feature | |

## Related Issue
#119

Issue #45, opened 2020-06-03
Can bombsquad PC display stylish font and emojis correctly?


bseditor:
First of all it isn't a big concern; still something to consider. PC players are facing this problem because now most of players use stylish fonts in their profiles and use emojis before or after their names which isn't display correctly on PC and creates difficulty to read players name for PC players. Hopefully, this issue will get fixed.

efroemling:
This is a known issue currently in the Windows build. (And also the Linux build).
The game can draw basic characters itself using its bundled textures (basic English and Cyrillic alphabets) but for any unicode characters beyond that (including Emoji) it has to be wired up to have the OS draw them. I've done this wiring up for the Android build and for Apple OSs, but I still need to do it for Windows/Linux. 

I'll definitely get this done before shipping on Steam or anything like that. I also plan to open source most of the graphics engine on here when I have time, so perhaps someone else can help add that functionality in.

Loup-Garou911XD:
Closing as it's on Eric's todo list

PR #141, opened 2020-08-02
New meta-tag '# ba_meta export keyboard'

Dliwk:
A new feature that provides a meta tag for custom keyboards offered in #137.

- To change keymap/keyboard quick press space two times.
- Keyboard may have unlimited pages with some extra chars or letters.

A word about standard emojis (#123)... Currently only Android (and Apple) devices can render them (this mentioned in #45) so is using them in all platfoms correct? For example, I see following:
изображение

### DIFF
--- a/.idea/dictionaries/roman.xml
+++ b/.idea/dictionaries/roman.xml
@@ -1,0 +1,7 @@
+<component name="ProjectDictionaryState">
+  <dictionary name="roman">
+    <words>
+      <w>maxlen</w>
+    </words>
+  </dictionary>
+</component>
--- a/assets/.asset_manifest_public.json
+++ b/assets/.asset_manifest_public.json
@@ -1,4 +1,5 @@
 [
+ "ba_data/python/__pycache__/keyboards.cpython-38.opt-1.pyc",
  "ba_data/python/ba/__init__.py",
  "ba_data/python/ba/__pycache__/__init__.cpython-38.opt-1.pyc",
  "ba_data/python/ba/__pycache__/_account.cpython-38.opt-1.pyc",
@@ -28,6 +29,7 @@
  "ba_data/python/ba/__pycache__/_general.cpython-38.opt-1.pyc",
  "ba_data/python/ba/__pycache__/_hooks.cpython-38.opt-1.pyc",
  "ba_data/python/ba/__pycache__/_input.cpython-38.opt-1.pyc",
+ "ba_data/python/ba/__pycache__/_keyboard.cpython-38.opt-1.pyc",
  "ba_data/python/ba/__pycache__/_lang.cpython-38.opt-1.pyc",
  "ba_data/python/ba/__pycache__/_level.cpython-38.opt-1.pyc",
  "ba_data/python/ba/__pycache__/_lobby.cpython-38.opt-1.pyc",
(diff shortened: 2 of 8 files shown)


efroemling:
This is great!
Notes:
- In response to your question: most people playing the game will be on platforms like Android that *can* display full unicode such as Emoji, so I think it's ok to assume they exist and just have blank buttons otherwise. My goal is to get support added to all platforms eventually so this should not be an issue forever. Or if you feel like it, you can do something similar to the language pop-up which only displays some languages when full unicode is available. I just added a _ba.can_display_full_unicode() function if you'd like to go this route.
- It appears that there is no unicode page on the default keyboard now (at least looking at it from the Mac prefab build); is this intentional? Could you add a screenmessage() error if they tap that button and there are no pages available or something to that effect? Currently it just does nothing which feels slightly broken. If you'd like to add an error message entry to assets/src/ba_data/data/languages/english.json and use a `ba.Lstr(resource='yourErrorMessageText')` to display it then it'll get translated to all languages (eventually).
- Same deal with double-clicking space; by default there is only one keyboard so it feels slightly broken in that it just silently prints a space and then deletes it. Could you have it perhaps print a screen-message about which keyboard it is switching to or mention that there are none available?

Anyway, thanks for contributing this! Looking forward to pulling it in.

Dliwk:
Ok; can someone check emojis?

efroemling:
Thanks for the changes; everything looks good to me. (emojis aren't showing up because they aren't drawable in the current prefab builds, but I'll test them on Android/Mac BombSquad builds to make sure they work). I'll get resources added for your text strings.

My one thought is that it is a little ugly having the 'press double-space to switch keyboards' instruction visible at all times since the vast majority of people won't install extra keyboards. I might change it so that message only shows up if there is more than one keyboard present.

Thanks again for this!

Would you by chance want to add an entry on the [Knowledge Nuggets](https://github.com/efroemling/ballistica/wiki/Knowledge-Nuggets) page showing how to make a custom keyboard? (or an entire dedicated wiki page if you want). That might help get people using this functionality. Or if you feel like it is documented well enough in the source code for people to discover it then that works too 👍

Dliwk:
Uh-oh; I missed a little bit, could you move assets/src/ba_data/python/keyboards.py in bastd?

efroemling:
No worries; I'm in the process of doing that right now :).  I'm putting it in bastd/keyboard/englishkeyboard.py
I'm also modifying the code a bit in _get_keyboard() so that it supports nested module names such as bastd.keyboard.englishkeyboard.EnglishKeyboard and adding text resources for the strings so they get translated.
I should have a commit up soon...

Dliwk:
Great, thanks!

I finished [Knowledge Nuggets](https://github.com/efroemling/ballistica/wiki/Knowledge-Nuggets#hello-world-3-creating-a-new-keyboard) page, but I still don't really like it, I couldn't write much there. When you have some free time, can you check it out?

efroemling:
Ok I've pushed commit 0a625831ae7d4897ebd326b1be3717f4687682ea with my tweaks to the keyboard stuff. Let me know if anything looks odd to you.

And I think your knowledge-nuggets entry looks good! I just made some minor adjustments such as mentioning that the user might want to turn on 'Always Use Internal Keyboard' to be able to see the results easier.

Thanks again for doing this! Much appreciated 😁

PR #142, opened 2020-08-02
Added import and export button

Benefit-Zebra:
## Description

I have added a Export and Import button to watch UI.

IMG-20200802-WA0017

IMG-20200802-WA0019

On pressing Export Button, the selected replay gets saved in BombSquad/replays directory. 
On pressing Import Button, all .brp replays get imported which are in replays folder. 

## Type of Changes
|   | Type |

### DIFF
--- a/assets/src/ba_data/python/bastd/ui/watch.py
+++ b/assets/src/ba_data/python/bastd/ui/watch.py
@@ -23,6 +23,7 @@
 from __future__ import annotations
 
 import os
+import shutil
 from typing import TYPE_CHECKING, cast
 
 import _ba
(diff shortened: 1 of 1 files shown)


DrovGamedev:
hello there: https://github.com/DrovGamedev/BombSquad-Hardcore-Modpack/blob/master/mods/replays.py

Benefit-Zebra:
The function of your writteb and my written scipt are same @DrovGamedev.

efroemling:
This is a great idea. My only notes would be:
- Can you fix the CI checks? I'm happy to offer feedback if you get stuck.
- I'd like to have those buttons in the same place as watch/rename/delete. If this part is too complicated I can do that for you though. It might be best to have a 'more' pop-up menu at the bottom to encompass options such as these since there might not be enough space for 2 more buttons.

Benefit-Zebra:
@efroemling How this UI looks:

IMG-20200803-WA0018

Sorry, I have by mistakely deleted the branch in which I have created this pull request, so I can no longer commit to it.
Now I will have to create a new pull request.

Dliwk:
I think this UI is not good idea... It will only work well with small app UI.
изображение
"More" pup-up menu (you can use bastd.ui.confirm.ConfirmWindow as example) really can be best solution, or something like @DrovGamedev mod.

Benefit-Zebra:
I have shifted the buttons to side in order to get more space.

efroemling:
Thanks for taking the time to try the right-side approach, but I would prefer to try and have all buttons on the left side. That way we preserve as much space as possible for descriptive replay names. I think a few options would work:
- Making just enough room under the 3 left buttons for a 'more' popup menu (as has been mentioned)
- Turning the left section into a vertically scrollable list so we can add more buttons even if they don't all fit on screen at once.

A nice benefit of either of these options is it allows adding more functionality in the future without running out of space. Perhaps we could make a meta-tag to define additional actions that operate on replays so modders could go to town with all sorts of ideas...

Benefit-Zebra:
> * Turning the left section into a vertically scrollable list so we can add more buttons even if they don't all fit on screen at once. I think I might actually prefer this option; this way all actions would get equal treatment instead of some being abstracted under a menu and some not.

I too liked this idea.

PR #147, opened 2020-08-05
New Watch UI

Gengar-lab:
Steps

## Description
Added a scroll widget to replay options. 
Also added Import, Export and Duplicate Button. 

IMG-20200805-WA0002

I am once again sorry as I by mistakely deleted the branch from where I created #142 so I closed that pull request. 

## Type of Changes
|   | Type |


efroemling:
Looking great so far!
My one main note is that the UI breaks in 'large' and 'medium' modes. (as used on tablets and desktop computers).
If you go to ba/_ui.py line 98 you can temporarily override the UI to display at UIScale.MEDIUM or UIScale.LARGE sizes to test that. If you can adjust your math to be based on the window size/etc. instead of hard coded then it should work at all UIScale values.

Benefit-Zebra:
> If you go to ba/_ui.py line 98 you can temporarily override the UI to display at UIScale.MEDIUM or UIScale.LARGE sizes to test that.

I tried this by writing a plugin, but there is no change in the UI when I open BombSquad on my mobile.

efroemling:
> > If you go to ba/_ui.py line 98 you can temporarily override the UI to display at UIScale.MEDIUM or UIScale.LARGE sizes to test that.
> 
> I tried this by using plugin, but there is no change in the UI when I opened BombSquad on my mobile.

You'll see that code in the watch UI uses the value of ba.app.ui.uiscale to create the window at different sizes. So if you are setting that value in a plugin or anywhere else then you should see the value used there be different (throw some print() statements or screenmessage()s or whatnot to verify).

Benefit-Zebra:
> You can also just override it temporarily at the place where the watch ui reads it.

I liked this idea. Can you please help me out by giving a screenshot of how old watch ui looks in PC for UIscale.MEDIUM and UIscale.LARGE .

JoseANG3L:
Hi, I thought of a way to order those buttons without using a scroll, but shortening the names of each function.
Captura de pantalla (32)

JoseANG3L:
Here I modified the position and size of the buttons, the green text and MyReplays went up a bit.

SMALL Screen
Captura de pantalla (33)

JoseANG3L:
MEDIUM Screen:
Captura de pantalla (34)

JoseANG3L:
This is the file, but the texts are edited to look like the image, all languages need to be edited.
[https://github.com/JoseANG3L/BombSquad-1.5.x/blob/master/watch.py](url)

Benefit-Zebra:
@JoseANG3L This is a good idea, but I will wait for @efroemling answer what he suggests, going with your idea, or anything other.

JoseANG3L:
@efroemling @Benefit-Zebra 
Hi, update the watch file to share replays for BombSquad 1.5.27.

For the Android platform it is perfect.

But for Mac and Windows platforms you need a function to open the replays folder, a function the same as the one to show mods folder.

Download here:
https://github.com/JoseANG3L/BombSquad-1.5.x/commit/0a24a2608aba92c4c9e4dbd1c244828951ba1dfa

JoseANG3L:
SMALL Screen
Screenshot_2020-11-09-16-12-43-869_net froemling bombsquad

MEDIUM Screen
Captura de pantalla (37)

LARGE Screen
Captura de pantalla (38)

Benefit-Zebra:
Thanks @JoseANG3L For updating script. I am quite busy with my online education.

efroemling:
Hey folks; just checking in here. What's the status of this? I just took a look at it and I'm seeing only a 'Share' button added which opens my mods folder when pressed, but I don't see any replay copied there.

JoseANG3L:
That problem is fixed by adding a command similar to mods to open the replays folder. But that command is in a file that I can't edit.

Benefit-Zebra:
@efroemling The main problem was that on Android we can't access root files without rooting the device hence we had no access to .brp files. So the idea was to update the watch script in order to provide Android users those .brp files by copying them into the mods folder.

Whereas for PC users such as Windows users, the data files are present for them in the C: Drive and they can access them, hence easily copy and paste them those .brp files.

Benefit-Zebra:
@efroemling The main problem was that on Android we can't access root files without rooting the device hence we had no access to .brp files. So the idea was to update the watch script in order to provide Android users those .brp files by copying them into the mods folder.

Whereas for PC users such as Windows users, the data files are present for them in the C: Drive and they can access them, hence easily copy and paste them those .brp files.

Aslo I have no idea why this CI check linux is exiting with error code 2.

Benefit-Zebra:
We are now going to open Replay directory when on PC so that users can easily copy and paste .brp which they need.

efroemling:
I know I've let this PR sit for way too long (sorry), so wanted to share some thoughts:

As of the soon-to-be-released 1.7 build, most android users will no longer have easy access to the mods folder (due to increased security in recent android versions which I might elaborate on elsewhere). I'm working on 'workspaces' which will be a platform-independent way to add scripts to devices, but they won't be usable for pulling files off of it. But in general I want to start discouraging things from relying on the mods folder since it is not universally (or at least not easily) accessible.

What I propose:
- I think Ballistica should have generic 'import-user-file' and 'export-user-file' functionality. These would hook into platform-specific functionality like file selection dialogs or share-sheets.
- This UI could then talk to that functionality. The 'import' button would be visible only if 'import-user-file' is available on the running platform and a 'share' button would only be available if 'export-user-file' is available.

The biggest bit of work here would be implementing the 'import-user-file' and 'export-user-file' system. If anyone would like to try that, let's chat. Otherwise I can try to get to it at some point, but it's not too high on my priority list at the moment.

Gengar-lab:
> What I propose:
> 
> * I think Ballistica should have generic 'import-user-file' and 'export-user-file' functionality. These would hook into platform-specific functionality like file selection dialogs or share-sheets.
> * This UI could then talk to that functionality. The 'import' button would be visible only if 'import-user-file' is available on the running platform and a 'share' button would only be available if 'export-user-file' is available.

Or simply the app could use `Android/data` folder, which I think the app would still have access to in recent android versions and can be used to export replays, eg. `Android/data/net.froemling.bombsquad/replays/myreplay.brp` and then the user can copy the replay.

But the problem would be importing replay as **FileSelectorWindow** would probably not have access to Android/data/net.froemling.bombsquad, though needs to be tested first 🤔.  

We could simply save each replay in the `net.froemling.bombsquad/replays` folder but that could I guess that could slightly affect performance as ___lastReplay.brp_ would be constantly overwriting after each game.

efroemling:
A bit more on this (after yet another delay; sorry).

As you mention, there are still places where Android apps can directly read/write data where the user has access to it in newer Android versions. But as I understand (correct me if I'm wrong) this data is blown away if the app is uninstalled, which is a bit of a bummer. I'd like for people to be able to uninstall/reinstall the app if something gets borked without it blowing away something that is valuable to them and possibly irreplacable. That’s part of the reason in general I'm wanting to move things towards being more tied to accounts; not devices or os-specific filesystem locations. (hence the concept of workspaces in 1.7 and asset-packages coming up in 1.8, as well as the eventual deprecation of device-accounts). So I think something like you're touching on in points 3 and 4 is the way forward. Basically having replays live in the cloud.

So, with that in mind, I still think the cross-platform ‘file-import-ui’ and ‘file-export-ui’ bits I was mentioning earlier would be useful additions to the engine.  We could wire those up to pull in or spit out replay files from their current internal locations, and then, if/when replays move to being cloud based, those same UIs could still function. Nice and clean and consistent across all platforms.

So I think for the sake of cleanliness I’m going to close this PR for now, since it is effectively blocked by that feature. If we can get that in place I’d be happy to re-open this and we can wire this up to use it.

And always happy to discuss any of this further or help spec out that feature if anyone wants to take a swing at it. I’ll try to get to it eventually, though 1.7 stuff is at the top of my priority list currently. And apologies for letting this sit so long; hopefully it will not be wasted effort.

PR #160, opened 2020-08-11
Added Rescan plugins Feature

Benefit-Zebra:
## Steps

## Description
It really gets annoying when you have to everytime restart your game when you enable/disable plugin or add a new plugin to folder when the game is on. 

So I have here added a Rescan Plugin button to Plugins Window which will helps us to enable/disable plugin or detect new plugin without restarting the Game. 

According to me, its a very useful feature since in BombSquad 1.4 or older similarly like the plugin case, to compile mods, we needed to restart the game. Thus this will save time of user. 

## Type of Changes
|   | Type |

## Example
```python

    # ba_meta require api 6

    from __future__ import annotations

    from typing import TYPE_CHECKING

    import ba
    from Example import Sample

    if TYPE_CHECKING:
        pass
    
    # Below line is necessary since it helps to restore 
    # the original state when disabling plugin
    copy1 = Example.Sample 

    class NewSample:
    - - snippet - - 

    # ba_meta export plugin
    class MyPlugin(ba.Plugin):
        """My first ballistica plugin!"""

        def on_app_launch(self) -> None:
            Example.Sample = NewSample

       def on_disable(self) -> None:
            Example.Sample = copy1 # Restoring Original State

    ```

### DIFF
--- a/assets/src/ba_data/python/ba/_plugin.py
+++ b/assets/src/ba_data/python/ba/_plugin.py
@@ -58,3 +58,6 @@ class Plugin:
 
     def on_app_launch(self) -> None:
         """Called when the app is being launched."""
+
+    def on_disable(self) -> None:
+        """Called when the plugin is disabled."""
--- a/assets/src/ba_data/python/bastd/ui/settings/plugins.py
+++ b/assets/src/ba_data/python/bastd/ui/settings/plugins.py
@@ -24,10 +24,11 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
+import _ba
 import ba
 
 if TYPE_CHECKING:
-    from typing import Tuple, Optional, Dict
+    from typing import Optional, Dict, Tuple, Callable, List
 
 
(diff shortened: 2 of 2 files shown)


efroemling:
This seems like useful functionality, but I'd encourage rethinking it a bit.
I would like to avoid doing things like re-running meta-scan stuff or firing on_app_launch() code at arbitrary times. Plugins will generally assume that those things happen in a set order and exactly one time, and if that is sometimes not the case I worry it would introduce subtle bugs since people would not be testing those cases. It is sort of like reloading a python module; it is technically possible, but in my experience the results are not always safe or predictable (though it can still sometimes be useful during development).

I think something like this could be great implemented as some sort of GUI plugin (perhaps with a warning that it may cause unpredictable results). I'm hoping to add a 'debug panel' GUI soon which is accessible from anywhere; perhaps something like this could go there..

efroemling:
I'm going to close this for now, though I do think it could be good to revisit once it is possible to write debug-gui plugins for this sort or purpose..

PR #164, opened 2020-08-14
CTF Flash bug

Benefit-Zebra:
## Steps

- [x] Ensure `make preflight` completes successfully..

## Description
When a player grabs the flag of enemy team, the base of enemy team flashes. Now if the player goes above the enemy's base carry the flag after 3 seconds, the base again flashes. You can see that in the video. 

[CTF Video](https://drive.google.com/file/d/19jk5dxxZqy5iFyXG7SQ7dynBUVsnFMiT/view?usp=drivesdk)

## Type of Changes
|   | Type |
| ✓  | :bug: Bug fix  |

### DIFF
--- a/assets/src/ba_data/python/bastd/game/capturetheflag.py
+++ b/assets/src/ba_data/python/bastd/game/capturetheflag.py
@@ -96,6 +96,7 @@ class Team(ba.Team[Player]):
         self.score = 0
         self.flag_return_touches = 0
         self.home_flag_at_base = True
+        self.flash_once = False
         self.touch_return_timer: Optional[ba.Timer] = None
         self.enemy_flag_at_base = False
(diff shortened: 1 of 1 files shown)


efroemling:
This behavior was intentional. If someone picks up a flag and is immediately stopped, I wanted it to flash again if someone comes in and picks up the same flag again a few seconds later. There is already a variable for controlling this length of time, so turning it up significantly is the same as just never having it flash a second time, making the code here redundant. Perhaps this would be worth opening a discussion about in the issues section if you find the default behavior undesirable. (perhaps we should increase the time, make it configurable, etc?..)

Benefit-Zebra:
Yeah I agree, but this should not happen if the same individual while holding flag goes up and down the flag spawn location and the location flashes.

PR #220, opened 2020-12-22
Meta subsytem now enables new plugin by default in headless builds

Dliwk:
## Description
Meta subsytem now enables new plugin by default in headless builds to allow server admins to get their modified server working 'out-of-the-box', without manually updating the config.

## Type of Changes


efroemling:
Good idea. Thanks!

PR #252, opened 2021-03-30
Shuffle FFA spawn positions

Dliwk:
## Description
Currently, if a fixed set of players play for a long time on the same map, they always spawn in the same positions, I don't think it's fun.

## Type of Changes
|   | Type |

### DIFF
--- a/assets/src/ba_data/python/ba/_map.py
+++ b/assets/src/ba_data/python/ba/_map.py
@@ -251,6 +251,9 @@ class Map(Actor):
                              or [(0, 0, 0, 0, 0, 0)])
         self.ffa_spawn_points = (self.get_def_points('ffa_spawn')
                                  or [(0, 0, 0, 0, 0, 0)])
+        # Let's shuffle our spawn points,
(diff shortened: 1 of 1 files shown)


efroemling:
Apologies for the slow followup on this.
I think randomizing the ffa spawn points used by players is a good idea, however I feel like this is not quite the right way to do it. There could be code unrelated to spawning that looks at a map's ffa spawn points and if they change unpredictably that could cause problems.
Perhaps simply setting _next_ffa_start_index to a random value instead of 0 would have the same effect?

Dliwk:
Okay, in principle, cyclically move the players will also go.

efroemling:
Thanks for the quick turnaround. If this winds up feeling too predictable we can generate a separate 'spawn-order' list which is a shuffled version of ffa-spawns and use that instead. I just think its a good idea to keep the ffa-spawn list constant.

efroemling:
...though I remember the ffa-spawn-point selection code tries a few points to get one as far from other players as possible, so  maybe purely random would be ok 🤔

PR #273, opened 2021-05-09
 Set winner when player leaves elimination mid-game

ritiek:
## Steps

- [x] Write a good description on what the PR does.

## Description
Some players immediately leave the game mid-way in elimination maps
when they have the final turn in the team and they are about to
die/fall off a cliff, which causes the game to draw when the opponent
team should have been declared the winner.

The game will now check if the player to leave the elimination game was
the last player in the team and in case they were so, mark their team
score so the opponent team is declared as the winner.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+- Declare opponent team as the winner if a player with their final turn leaves an elimination game.
+
 ### 1.6.0 (20357)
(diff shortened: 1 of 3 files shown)


efroemling:
Looks good; thanks!

PR #302, opened 2021-07-10
Co-op in headless builds

Dliwk:
## Description
Stuff discussed in #278. Sadly but I couldn't handle playlist selection UI :( However now it allows to run co-op levels on servers.

Don't merge for now, this opened for discussions; I'll refine it.

Troubles:
- If player all player leaves, activity will end and show Score Screen, but no one of joined players can continue the game. `on_player_join` haven't called (really haven't, not just invisible print) for them :thinking: Eww probably I'll anyway have to remove this stuff because now (and why only now?) I see errors in the console...

## Type of Changes

|   | Type |

## Screenshots 
изображение

### DIFF
--- a/.idea/dictionaries/roman.xml
+++ b/.idea/dictionaries/roman.xml
@@ -1,6 +1,7 @@
 <component name="ProjectDictionaryState">
   <dictionary name="roman">
     <words>
+      <w>gamename</w>
       <w>maxlen</w>
       <w>pagename</w>
     </words>
--- a/assets/src/ba_data/python/ba/_coopsession.py
+++ b/assets/src/ba_data/python/ba/_coopsession.py
@@ -158,7 +158,6 @@ class CoopSession(Session):
 
         if _ba.app.server is not None:
             # If we're in server mode, end game and show results.
-            activity = self.getactivity()
             _ba.timer(2.0, WeakCall(self._end_activity_if_empty))
         else:
             # Otherwise, if all our players leave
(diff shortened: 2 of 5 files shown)


Dliwk:
I fixed this trouble:
> * If player all player leaves, activity will end and show Score Screen, but no one of joined players can continue the game. `on_player_join` haven't called (really haven't, not just invisible print) for them thinking Eww probably I'll anyway have to remove this stuff because now (and why only now?) I see errors in the console...

Actually this needs some cleanup form internal layer I guess. E.g. I suggest set 'enableContinues' account misc val to False for Server* accounts, this cause pauses after lose (when game offers server to spend some tickets xD).

Well, I kind of tested this, everything seems to be working now.

Dliwk:
Oops, I forgot about this. The reason was following behavior: when mid-activity joins disallowed in session level, people who join in ScoreScreenActivity were kind of invisible for activity and it couldn't use `on_player_join` stuff to bind players' inputs to start game.

I don't mind discussing this moving right here either. Have to say that this moving isn't finished, how it's probably visible and what I forgot about.

efroemling:
I did a bit of tidying in commit 6ff2b3093332a1c017a9c8df954f09975b47c621 if you'd like to take a look. I tried to unify the logic in _coopsession where it can leave on empty sessions and I changed the server config from a single 'coop_game_name' to 'coop_campaign' and 'coop_level'. Let me know if anything looks bad to you or you see any problems.

efroemling:
> I didn't think that would make so much sense for games which allow mid-activity joins. I remember many cases where all players left only to change her character and join back to continue playing. 

Yes games that allow mid-activity-joins should continue to function as they do now; I just mean we should check games that *don't* allow mid-game joiners to make sure they still behave correctly. They should all already contain logic to end themselves if there are no players left in them; I just want to make sure that logic is still working. Just something to watch out for. And in the future it might make sense for the Session to automatically end games in that situation instead of relying on each game to do it itself.

Issue #151, opened 2020-08-06
Spaz Bigger than it should be


PythonCode-blip:
This was a bug that was on earlier versions of the game and somehow it came back, it also makes the spaz bigger just by jumping repeatedly (doesn't work in Happy Thoughts) and giving you an advantage in battle.
![Screenshot_2020-08-05-19-54-40](https://user-images.githubusercontent.com/66983599/89485518-d0977400-d755-11ea-98ca-456b790bfb0c.png)

TheMikirog:
First of all, that's not a bug - it's a quirk of the engine and how jumping works. You extend legs to jump and spamming the button results in that.

Second, this doesn't give you an advantage in battle at all. Yes, hitting the head is difficult, but you move ridiculously slowly and for a game about speed, gimping yourself like that is not worth it.

Why are we talking about this?

YairMejiaValtierra:
> First of all, that's not a bug - it's a quirk of the engine and how jumping works. You extend legs to jump and spamming the button results in that.
> 
> Second, this doesn't give you an advantage in battle at all. Yes, hitting the head is difficult, but you move ridiculously slowly and for a game about speed, gimping yourself like that is not worth it.
> 
> Why are we talking about this?

but bro u dont understand in 1.5 servers asia that does give an advantage because when u jump and the spaz at 60° angle to left or right and the swipe or joystifk is at the same velocity of where u go the jump is bigger than bombjump and u can literally just punch a spaz head off and giving u an insane amounted spead but this never happrn in 1.4 it was a weak punch before and it kinda gave u an adbantage it will still give u -100% dmg but in 1.5 it always give u over -180% dmg so this is too op and it ruins my gameplay for players spaz users it is annoying it will make players mad anf quit

YairMejiaValtierra:
Also why im saying Asia? is because that is the majority of where it happens same with SudAmerica servers

Dliwk:
I think TheMikrog is right (why the funny game feature is a bug? I would call it bug if the spaz then flew up, the sky turned green and the map fell to pieces), so I'll close this. 

Reopen if you don't agree and want to discuss.

PR #341, opened 2021-11-19
Restore old jump behaviour

### LINKED ISSUE #230: Discussion
I'm Yair Mejia Valtierra in a different account, still bothered about the long legged spaz and even if it isn't a bug, it should still be dealt with, why? because it gives you advantages in 1.4 servers meaning that you jump at an angle and wait for the spaz to be in a 70 degree angle to the left or right depending on the position, and when you're gonna touch the ground, you jump spin and punch giving you extreme momentum and giving you 200 punch power, you need to fix this, this is more annoying than bombjump and it made players leave game.

ritiek:
## Steps

## Description

Restores v1.4 jump behaviour. Means players can no longer get a long-legged spaz by spamming jump.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Related Issue

Closes #230, #151.


ritiek:
No problem! I was more afraid of people getting used to the new jump behavior. I rebased my branch.

efroemling:
Thanks; I'll pull it in now.
Appreciate it. I hope to eventually add a fix for this at the C++ layer (which is why I originally pulled it from the Python layer) but it is probably best to keep it present in the Python layer until that actually happens. 👍

PR #347, opened 2021-12-22
Fix SharedObjects

IvanPragma:
## Description
After changing the storage location of the `SharedObjects` instance (bc22359dcd06b1dea6a1c767ba350783f4fb53f5), you forgot to change the check in `SharedObjects.__init__`

## Type of Changes


efroemling:
Good catch; thanks for the fix!

Issue #368, opened 2022-03-07
Gameplay Issues


Dliwk:
I decided to collect everything here that seems wrong to me and potentially could be fixed, but them all in some way relates to the gameplay, so this probably needs to be discussed with everyone here. 

Anyone who has some thoughts, please post them here.

Here is the list (it's updated if something resolves):

- [ ] No "Player Scores!" announcement in CTF games.
- [ ] Should dead players win the race?
- [ ] Should dead players interact with flags?
- [ ] Chosen One: No collision when flag spawns right in place player stands.
- [ ] Blocking player in a chat.

#### No "Player Scores!" announcement in CTF games.
Similar to Chosen One issue;
![Безымянный2](https://user-images.githubusercontent.com/47295209/157112327-7ce3e42b-635b-457b-84cc-a10d0d9b8ff1.gif)

#### Should dead players win the race?
![Безымянный(1)](https://user-images.githubusercontent.com/47295209/157110256-0b600d19-ac5c-47e2-9b91-ddfbd85f9bff.gif)
In principle, it's cool, but suddenly everyone around it is unrealistically annoying :man_shrugging: 

#### Should dead players interact with flags?
Same as above, but now the flag acts as the finish line and teleports to base when touches a calmly lying dead body (and that is sometimes annoying unlike the one above).

#### Chosen One: No collision when flag spawns right in place player stands.
![изображение](https://user-images.githubusercontent.com/47295209/157105969-4453efec-aa3b-479a-9df2-19ccc7593d3d.png)

#### Blocking player in a chat.
Host mutes every client after 5 quickly-sent messages, but 5th message never shown.

![изображение](https://user-images.githubusercontent.com/47295209/157104717-e4af1418-b5e0-4f58-adb9-3afdfc6ffda2.png)
Sent messages: 'Hey', 'I do', 'not', 'spam', ~~'!!!'~~.

DavidReadsGitHub:
Can one of the player’s “splitted” corpses even touch and win the race?

Dliwk:
> Can one of the player’s “splitted” corpses even touch and win the race?

Is this a question or a topic for discussion? If first, then yeah, I've seen this before.

efroemling:
Chiming in:
Most of these seem like legit bugs to me and I'd be happy to accept PRs if anyone wants to fix them. A few notes:

[...]

PR #417, opened 2022-06-03
Fixed issues from #368 

Drooopyyy:
## Steps

- [x] Write a good description on what the PR does.

## Type of Changes


Drooopyyy:
-Dead players cant win race.
-Dead players cant interact with flags.

efroemling:
Looks good; thanks! Can you get the checks fixed and add a changelog entry? Let me know if you get stuck on anything.

Drooopyyy:
Thank you dliwk for helping me and fixing checks :) .

Dliwk:
Looks like some logic mess in PythonClassNode::GetDelegate() which can return None even if doraise=True ([line 206][1])

[1]: https://github.com/efroemling/ballistica/blob/c3348afb404d998708d494b8963928390ef0afa6/src/ballistica/python/class/python_class_node.cc#L206

Dliwk:
Ok I fixed a few things and now it works fine.

Dliwk:
Didn't found how can I increment app build number so I added changelog entry under 1.7.1 :man_shrugging:

efroemling:
> Didn't found how can I increment app build number so I added changelog entry under 1.7.1 🤷‍♂️

That works. I can move things to the 1.7.2 section if there are no more 1.7.1 builds (which is probably what I'll do).

Currently I have a script on the internal repo which handles changing build-number/version since there's a number of files across platforms that need to be changed. Maybe I can expose that but I'd need to make sure there's no weirdness with the internal lib thinking its one build number and the public repo bits thinking its another. Or maybe the internal lib should have a separate build number.

efroemling:
> Looks like some logic mess in PythonClassNode::GetDelegate() which can return None even if doraise=True ([line 206](https://github.com/efroemling/ballistica/blob/c3348afb404d998708d494b8963928390ef0afa6/src/ballistica/python/class/python_class_node.cc#L206))

Oh wow yeah I think you're right. Good catch! Looks like it would be possible for the type checker to think you have a non-optional delegate type after the getdelegate() call but in reality you have None.
Would you want to add that fix to this PR?

Dliwk:
> Oh wow yeah I think you're right. Good catch! Looks like it would be possible for the type checker to think you have a non-optional delegate type after the getdelegate() call but in reality you have None. Would you want to add that fix to this PR?

I'm tired a bit now, so I can add that tomorrow or you can do that yourself. And not sure where else could this be the case.

efroemling:
> > Oh wow yeah I think you're right. Good catch! Looks like it would be possible for the type checker to think you have a non-optional delegate type after the getdelegate() call but in reality you have None. Would you want to add that fix to this PR?
> 
> I'm tired a bit now, so I can add that tomorrow or you can do that yourself. And not sure where else could this be the case.

Ok I'll try to get to it today. And yeah it could be a good idea to scan over similar code patterns where a 'doraise' is present to make sure we're not violating type promises anywhere else..

efroemling:
> I'm tired a bit now, so I can add that tomorrow or you can do that yourself. And not sure where else could this be the case.

Ok I just added a fix for this in c3348afb404d998708d494b8963928390ef0afa6
I also looked through other calls with 'doraise' in the C layer and didn't see any other occurrences, but let's keep our eyes open for that sort of thing in the future. Thanks again for the heads-up.

And I'll go ahead and merge this now. Thanks @Drooopyyy for the fixes!

PR #419, opened 2022-06-04
chatmessage to specific client fix

### LINKED ISSUE #342: Chat messages for a specific player do not work
A long time ago (in version 1.5) I tried to send a message to a specific player, but I failed. Nothing works on 1.6 either.
When I set `clients`, no message is sent to the selected player.
#32

imayushsaini:
## Description
For some reasons pointer to Python::GetPyInts(clients_obj) is not working and we get empty (not nullptr) vector  in connection_set.SendChatMessage(), hence when we pass clients in _ba.chatmessage  it dont match with any client.

## Type of Changes

## Related Issue
Closes #342


efroemling:
First off, thanks @imayushsaini for looking into this and for submitting the fix.

I always like to know exactly why things are failing (so I can be sure a fix isn't just working by through blind luck) so I dug around a bit and figured out what the problem was:

This line pulls Python ints into a std::vector:
https://github.com/efroemling/ballistica/blob/7a7db61f29ca1d52d749f900fd49337bb15ceb4b/src/ballistica/python/methods/python_methods_ui.cc#L2058

And then the next line grabs a pointer to that vector for passing to SendChatMessage().

...the problem is that the `clients` vector is declared under the `if` scope 2 lines up, so it gets destructed one line later after we exit that scope. So when we call SendChatMessage we're holding a pointer to memory that is no longer valid. Oops. I guess that memory just happened to stay intact while I was testing it so I didn't notice.

While your fix does correct the issue by replacing pointer use with references, it also changes the SendChatMessage() behavior to take 'empty list' to mean 'send to everyone'. So _ba.chatmessage('HELLO', clients=[]) would be the same as _ba.chatmessage('HELLO') which is not ideal. The reason I set it up to use pointers was so nullptr could be a special 'send to everyone' value distinct from an empty list which would result in a send to no-one.

So I just pushed commit f06a380758f255c33b4eb659d6f694e8c3cc5d02 in which I fixed it by moving the `clients` vector out of the `if` scope. This means it remains valid until the end of the function so passing a pointer to it to SendChatMessage() will now behave.

Thanks a ton for tracking this down; that bug has been in there for quite a while! 😬

PR #421, opened 2022-06-05
Update getcurrency.py

itsre3:
## Description

Fixed tiny mistake in getcurrency.py
Screenshot_20220605-164433


efroemling:
Looks like I broke this during a routine tidying pass about 11 months ago. Thanks for the fix!

PR #429, opened 2022-06-16
Own PlayerSpaz

Easy10781:
## Description
Allows you to add your own PlayerSpaz types

Example:
```python
# ba_meta require api 6

import ba
from bastd.actor.playerspaz import PlayerSpaz


class MyPlayerSpaz(PlayerSpaz):
    def on_jump_press(self) -> None:
        ba.screenmessage("Hello from my playerspaz!")


class Player(ba.Player['Team']):
    playerspaztype = MyPlayerSpaz


class Team(ba.Team[Player]):
    pass


# ba_meta export game
class MyGame(ba.TeamGameActivity[Player, Team]):
    name = "Example"
```

## Type of Changes

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -32,3 +32,6 @@
 
 ### Droopy
 - Fixes in some minigames
+
+### Easy10781
+- Added feature
--- a/assets/src/ba_data/python/ba/_gameactivity.py
+++ b/assets/src/ba_data/python/ba/_gameactivity.py
@@ -853,7 +853,7 @@ class GameActivity(Activity[PlayerType, TeamType]):
         color = player.color
         highlight = player.highlight
 
-        playerspaztype = getattr(player, 'playerspaztype', None)
+        playerspaztype = getattr(player, 'playerspaztype', PlayerSpaz)
         if not issubclass(playerspaztype, PlayerSpaz):
             playerspaztype = PlayerSpaz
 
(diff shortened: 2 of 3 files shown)


efroemling:
Thanks for the contribution!
I do support the idea of being able to customize this; I just have one note:

I've got plans (hopefully for 1.8?) to revamp how spawning works so that it will be possible to spawn other actor types for Players; not only Spaz related ones. So with that future in mind of that I'd like to avoid importing anything Spaz-related in _player.py.

But in the meantime I think it doesn't hurt anything to expose this functionality, even if it is treated a bit as a temporary hack until that full generalization is done. But my thought would be to keep it limited to _gameactivity.py. How about something like this:
```python
# Use a custom spaz type if this Player has one; otherwise go with standard.
playerspaztype = getattr(player, 'playerspaztype', None)
if not issubclass(playerspaztype, PlayerSpaz):
    playerspaztype = PlayerSpaz
```

I think that'll still do what you're aiming for, right?

Easy10781:
yes, I guess I'll do that, and since ```python getattr(player, 'playerspaztype', None)``` is used here, I'll remove a that ugly code 
```python
try:
    self.playerspaztype = getattr(self, 'playerspaztype')
except AttributeError:
    self.playerspaztype = None
```

efroemling:
Ok I'll pull this in now. Thanks again for the contribution!
The one change I'll make is to remove the remaining mentions of PlayerSpaz from the _player.py for the reasons I mentioned before (keeping this a bit 'unofficial' since upcoming functionality will replace it). That is just an annotation though and should not interfere with the functionality; you can still assign a playerspaztype on a Player class and the getattr() will pick it up.

PR #436, opened 2022-06-24
Update kingofthehill.py

itsre3:
Fixed small issue


efroemling:
Interesting; looks like this got accidentally changed from `opposingnode` to `sourcenode` in a PR a few weeks back and I didn't catch it. Good catch!

PR #458, opened 2022-07-07
Update trophies.py

itsre3:

This should solve this 
Screenshot_20220707-094249


efroemling:
Thanks for reporting this! I hadn't noticed it was broken.
However I think I will fix it in a slightly simpler way.
The _create_trophy_type_widgets() function is supposed to return the total points for the trophies which should work fine without needing to assign anything to self. The problem seems to be that its variable for the total number of points and the variable holding the points for an individual trophy type each time through the loop are both `pts`. I'm guessing they used to have different names and the bug was introduced when I accidentally renamed them with the same name. This means that each time through the loop replaces the total count with whatever that line was.
So I just changed those variables to `total_pts` and `this_pts` and it should be behaving now. (commit adb979cc867fb0fe9f7115d457779e535987059d)
So I'll go ahead and close this, but thanks for tracking this down!

PR #460, opened 2022-07-08
Revert _campaign.py

vishal332008:
I found this bug in 1.7.4.

The first game is locked. (Onslaught training)

BombSquad 08-07-2022 18_37_29

If i click it , then it tells to complete Uber Runaround. (In easy mode , i finished till uber runaround , so i guess it is showing me that name , it might differ with others , i didn't check it yet)

BombSquad 08-07-2022 18_37_45

I hope that this gets fixed.

### DIFF
--- a/assets/src/ba_data/python/ba/_campaign.py
+++ b/assets/src/ba_data/python/ba/_campaign.py
@@ -49,12 +49,12 @@ class Campaign:
         """Whether this Campaign's levels must be played in sequence."""
         return self._sequential
 
-    def addlevel(self, level: ba.Level, index: int = -1) -> None:
+    def addlevel(self, level: ba.Level) -> None:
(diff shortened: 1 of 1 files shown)


efroemling:
Ah.. whoops. My bad! Good catch!

I went ahead and pushed a fix in commit 80a132ec7da7b80c51866a0f1af76590758730a5 which uses None to specify regular at-end-of-list addition, so I'll go ahead and close this.

I always try to avoid using large numbers like 99999 for those sorts of things when possible; it just makes me feel a bit dirty 😬.

I recently read about some code that long ago started using a time-out value like 999999 to mean 'this thing should never time-out' but recently that offset started causing times to pass the 2038 unix max 32 bit time value and wrap around to the past and so all their stuff broke without warning. 🙈

PR #483, opened 2022-08-05
Weak punch sound to reuse (spazfactory.py)

Ralfreengz:
I know this change is very small and unnecessary, but I saw this sound in very old versions of bombsquad and said why not bring it back?


efroemling:
(also this one needs updating too)

PR #488, opened 2022-08-14
Updated ChargerBotPro class

CrossJoy:
Deleted  "default_shields = True" line from the class. 

## Steps

- [x] Write a good description of what the PR does.

## Description

Edited the spazbot.py in bastd/actor file. Remove shield enable from ChargerBotPro to distinguish between ChargerBotPro and ChargerBotProShielded.

## Type of Changes

|       | :sparkles: New feature |
|       | :hammer: Refactoring   |
|       | :scroll: Docs          |

### DIFF
--- a/assets/src/ba_data/python/bastd/actor/spazbot.py
+++ b/assets/src/ba_data/python/bastd/actor/spazbot.py
@@ -733,7 +733,6 @@ class ChargerBotPro(ChargerBot):
     """
     color = PRO_BOT_COLOR
     highlight = PRO_BOT_HIGHLIGHT
-    default_shields = True
     default_boxing_gloves = True
     points_mult = 3
 


efroemling:
Thanks for the PR! I see that this removes a redundant class attribute assignment. My one note would be to mention the reason for removing it in the future (the fact that it is redundant). Otherwise it might look like it is changing gameplay behavior which opens it up to more scrutiny.

efroemling:
Actually I misspoke in my previous (deleted) comment. I thought this was removing a redundant attribute assignment from ChargerBotProShielded but it is actually removing it from ChargerBotPro. While that does make the class names more correct, it also changes gameplay which I want to avoid. I don't remember exactly why at this point, but while calibrating gameplay I must have added shields to ChargerBotPro for balance purposes. So I'd rather not remove them now for fear of throwing off existing high scores/etc.
I'd be down to come up with a more accurate name for ChargerBotProShielded, but even that might cause more trouble than its worth by breaking people's mods.

PR #489, opened 2022-08-16
Nerfed pickup cooldown a bit

MrJetzzzz:
## Steps

## Description
To avoid pickup abuses

## Type of Changes

## Related Issue
Many players spamming the pickup button to make his/her victim fall to the map's bounds while avoiding being punched by the victim


vishal332008:
My suggestion , just become Pro in the game. 🙂

EraOSBeta:
Hmm, pickup cooldown seems nice, but can you explain the system behind it? Is the cooldown always active? Or does it only affect if you spam pickup? How much is the cooldown? And does it effectively stop the spam and make it fair for all players?

EraOSBeta:
> Hmm, pickup cooldown seems nice, but can you explain the system behind it? Is the cooldown always active? Or does it only affect if you spam pickup? How much is the cooldown? And does it effectively stop the spam and make it fair for all players?

Ok, I take that back, it just adds a cooldown to all grabs, that sounds like a terrible idea

TheMikirog:
Honestly, this system could've been implemented with more care and attention. This edit tries to do a quick and dirty fix that I don't think was through through or playtested well enough.

The issue is that misses also count for the cooldown, which feels terrible to play.
JRMP did this and the animations aren't distinct enough to see when grabs miss, since there's no feedback, visual or auditory.

What I ended up doing in my mod instead is only applying the cooldown once you release the victim. This ensures that you have full freedom on how you throw people, but if you fail on that throw, you deserve to get punished. Should I implement that here?

efroemling:
Ok it sounds like this idea warrants some further discussion; maybe the cooldown-on-release idea or something else. We can chat about it in Discord (or open a discussion here). Closing this in the meantime. Thanks for getting the ball rolling with this submit though.

PR #490, opened 2022-08-16
Update spaz.py

itsre3:
## Steps

## Description

Default boxing gloves seems to be override when gloves powerup is picked
This should solve it to some point


efroemling:
Good catch; thanks!
Side-note: I'm wondering if there's a similar issue with default_shields
(I suppose I never considered either case since those I only used those values for bots which don't pickup powerups)

PR #495, opened 2022-08-26
FreezeMessage handling improvements

Dliwk:
## Steps

- [ ] Add a CHANGELOG entry if your change is non-trivial. (is it trivial?)

## Description
Currently if Spaz is already frozen, it ignores any incoming `FreezeMessage`s (so if spaz is about to thaw, but not actually did it yet, ice bomb explosion won't have much effect, unlike when spaz is just thawed). This PR changes this behaviour so that new messages extend this effect up to 5 seconds.

## Type of Changes

### DIFF
--- a/assets/src/ba_data/python/bastd/actor/spaz.py
+++ b/assets/src/ba_data/python/bastd/actor/spaz.py
@@ -206,6 +206,7 @@ class Spaz(ba.Actor):
         self._turbo_filter_counts: dict[str, int] = {}
         self.frozen = False
         self.shattered = False
+        self._thaw_timer: ba.Timer | None = None
         self._last_hit_time: int | None = None
         self._num_times_hit = 0
         self._bomb_held = False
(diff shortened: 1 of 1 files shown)


efroemling:
This was definitely a design decision. I know it seems logical that freezing someone who is frozen would extend the freeze, but I thought it would make it too easy to completely kill someone by hitting them with freeze bombs repeatedly. This way you at least have to time it perfectly to re-freeze them just after they thaw.

I'll leave this open for now if people disagree or want to comment. And this also makes me think about if there would be a good time to make changes such as this to minimize the effect on high scores/etc. (for instance, maybe we could make a list of changes we want to make and then apply them all for SceneV2 or the 1.8 release or something).

TheMikirog:
I agree with @efroemling on this one. Allowing the freezing duration to be extended opens up so many avenues from griefing potential, especially on maps like Step Right Up. 

If you don't like someone and you want to piss them off, just keep freezing them over and over again until one of you gets bored or your powerup duration ends.
The system we have now rewards proper timing and even if you get thawed and then frozen again, there's a chance the victim will fall and break themselves. The benefits of refreezing enemies this way are not as worth it as just punching them and getting it over with.

Dliwk:
Hmm yes that seems reasonable, but what about extending for a half of freeze time? third? Currently it's too sharp transition between 1 ms before thaw and 1 ms after thaw.

TheMikirog:
I think this would add ambiguity over how much time you're actually frozen.
When I freeze you the first time, I know exactly when you'll pop out of that state.
Now that I throw a bomb at you once more, I have no idea when that will happen.
And I can't learn that new time, because it changes depending if I freeze you again right away, if I freeze you when you're just about to thaw out and when I freeze you while you were about half-way through the timer.

I just don't see the benefits for this mechanic at all, no matter how much you want to split hairs over how much time extension is "good". You are trying to fix a problem that I don't think exists.

efroemling:
Thanks again for the submit, but yeah I do think that keeping the rules here as simple and understandable as possible is good for gameplay. I think its good that people can develop a sense of exactly when someone will un-freeze and it can make for exciting moments of success/failure if another freeze attempt goes off just before or just after that happens. It could be interesting to have a more 'soft' freeze effect where maybe it is weaker but its effects could stack or something, but perhaps that would be best as a different powerup.

PR #493, opened 2022-08-23
Removed Additional Time

vishal332008:
## Description
After you lose the rookie/pro/uber football game in campaign , the continue or end game window appears. The time that the player takes to choose and click an option (continue or end) also gets calculated into the football game time. 

|     | Type                   |
|-----|------------------------|
| ✓   | :bug: Bug fix          |

PR #497, opened 2022-08-29
Pause/Resume Current Activity

vishal332008:
Continuation of #493 

Thanks to @dliwk for the proper code.


efroemling:
Thanks for the fix!

PR #503, opened 2022-09-05
PB-ID Copy Button

### LINKED ISSUE #502: [Bug] PopupText 'offset' not working properly
When modifying BombSquad, the PopupText class has an error in the offset attribute, since it is not distributed correctly.
I attach the code that it should have.
The class is in the following location: bastd/actor/popuptext.py
Code line: 38-41

wrong

right

vishal332008:
New Feature and Closes #502

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -35,3 +35,7 @@
 
 ### Easy10781
 - Added feature
+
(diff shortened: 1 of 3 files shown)


efroemling:
Thanks again for doing this!

PR #506, opened 2022-09-08
Accept "fairydust" as an emit type in ba.emitfx()

ritiek:

## Description

For example, this now works:
```python
ba.emitfx(
    position=(10, 10, 0),
    velocity=(0, 2, 0),
    count=5,
    spread=0.1,
    emit_type="fairydust",
)
```

Seems to me it got missed but feel free to close this if there's a reason it wasn't done on purpose.

## Type of Changes


Dliwk:
Running `make update` should fix failing CI job.

efroemling:
Thanks for this!

ritiek:
@efroemling I think you will have to `make update` since it causing the ci to fail on master.

PR #514, opened 2022-09-22
Manage Account Button for V2 Accounts

vishal332008:


efroemling:
Thanks for doing this! I'll go ahead and pull it in now but just keep it disabled for the moment. I'd like to add a translation string for it and also add some back-end logic so it will let the browser access your account even if you're not already signed in there (which is how it will have to work for Google Play Games accounts since they *can't* sign in directly)

PR #515, opened 2022-09-23
Update spaz.py

vinnytherabbit:


assets/src/ba_data/python/bastd/actor/spaz.py
```diff
@@ -1052,21 +1052,21 @@ def handlemessage(self, msg: Any) -> Any:
                 self.node.handlemessage('flash')
 
                 # If we're holding something, drop it.
-                if damage > 0.0 and self.node.hold_node:
+                if damage >= 0.0 and self.node.hold_node:
```
Dliwk, line 1055: Could you also tell something about these `> 0` -> `>= 0` changes? Seems like this things (as `damage` and `self.hitpoints`) can't be negative, so that's always evaluated as True.

assets/src/ba_data/python/bastd/actor/spaz.py
```diff
@@ -1225,7 +1225,7 @@ def drop_bomb(self) -> stdbomb.Bomb | None:
         pos = self.node.position_forward
         vel = self.node.velocity
 
-        if self.land_mine_count > 0:
+        if self.land_mine_count >= 0:
```
efroemling, line 1228: This looks like it will completely break the logic. Just wanting to echo @Dliwk in asking what the motivation is behind these > to >= changes?...

PR #516, opened 2022-09-23
Update tournamententry.py

### LINKED ISSUE #74: Tournament UI remains on top after transferring to the store
(pasting from an email)
I found a problem with the tournament menu. It is nothing very serious, just something that is not working properly. When I click to use tickets in the tournament and I don't have enough, the game automatically warns me that I don't have tickets and a button to go to the store appears. If I click on the button, I am directed to the store, but the entry interface for the tournament remains on top.

itsre3:
Closes  #74


efroemling:
Thanks for the fix!

PR #521, opened 2022-09-29
Pop sound to reuse (spazfactory.py)

vinnytherabbit:
## Description
A sound to reuse.


efroemling:
Hi!
Thanks for the contribution, but I'm not sure this belongs here. The media loaded by SpazFactory is supposed to be things that are directly used when creating Spazzes. I don't think the pop sound is used by Spaz actors by default, is it?

efroemling:
A long time ago bombs would be spawned on the ground so I think the pop sound made more sense then to show that something happened. But now characters make a sound when whipping out a bomb so I think having the pop in addition just started to be too much and sound too chaotic in large games.

efroemling:
Regardless, thanks for the contribution :)

PR #524, opened 2022-10-08
changing device name for servers

imayushsaini:
party name as device name for headless 🤔.

### DIFF
--- a/assets/src/server/ballisticacore_server.py
+++ b/assets/src/server/ballisticacore_server.py
@@ -573,6 +573,11 @@ class ServerManagerApp:
         # slight behavior tweaks. Hmm; should this be an argument instead?
         os.environ['BA_SERVER_WRAPPER_MANAGED'] = '1'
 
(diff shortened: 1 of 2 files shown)


fraq-sss:
I support, good idea!

imayushsaini:
> I support, good idea! But that's not good if server name is long.

This name will appear only on ballistica.net/devices not ServerXXXX of game roster.
So long server name doesn't matter (can set limit on frontend if it breaks UI).

efroemling:
Looks good to me; thanks!

My one thought: would you mind if we went with BA_DEVICE_NAME instead of BA_HOST_NAME? Conceptually they are sort of interchangeable but internally the value is accessed via GetDeviceName() so I think this makes it more consistent.

(I'd be happy to pull this in and make that change on my end if you're ok with it)

efroemling:
Thanks again for doing this and for the quick turnaround on the tweak request!

PR #530, opened 2022-10-16
If we're knocked out, we'll drop any object that we're holding 

vinnytherabbit:
If we're knocked out and we're holding a object, we'll drop it immediately

### DIFF
--- a/src/ballistica/scene/node/spaz_node.cc
+++ b/src/ballistica/scene/node/spaz_node.cc
@@ -3884,7 +3884,7 @@ void SpazNode::Step() {
     if (!holding_something_ && hold_node_.exists()) hold_node_.Clear();
   }
 
-  // If we're knocked out, stop any mid-progress pick up.
+  // If we're knocked out, drop whatever we're holding.
   if (knockout_) {
     DropHeldObject();
   }


Dliwk:
That does look like something that might work (:tada:), but right now that's not true.

src/ballistica/scene/node/spaz_node.cc
```diff
@@ -3884,6 +3884,11 @@ void SpazNode::Step() {
     if (!holding_something_ && hold_node_.exists()) hold_node_.Clear();
   }
 
+  // If we're knocked out, stop any mid-progress pick up.
+  if (knockout_) {
+    holding_something_ = false;
```
Dliwk, line 3889: I've just got this working by replacing this line with 
```c++
DropHeldObject();
```
Think about whether it suits you.

Dliwk (approved):

IlScottlI:
Cannot wait for this update!

efroemling:
Thanks for the contribution!

This does look like a reasonable change. The only pitfall is that it affects the C++ level simulation, which gets run by all clients in a networked game - meaning that if some clients are running older versions of the game and some are running a newer version with this change, they'll see different behavior which will lead to ugly physics corrections and snapping for some people.

There are ways to deal with this but they're not especially simple (see the behavior_version attribute on the spaz node).

One of my big goals for the next update (1.8) is scene-versions which will let us split out all the c++ level sim stuff into different versions where we can make logic changes like this without worrying about breaking older versions, so I vote we sit on this change for now and try to get it in for 1.8.

PR #536, opened 2022-10-22
added ingame ping

imayushsaini:
## Type of Changes

## Description
Added ability to check current ping while playing in server.

image

### DIFF
--- a/assets/src/ba_data/python/bastd/ui/settings/advanced.py
+++ b/assets/src/ba_data/python/bastd/ui/settings/advanced.py
@@ -445,6 +445,17 @@ class AdvancedSettingsWindow(ba.Window):
         )
 
         v -= 42
+        self._show_game_ping_check_box = ConfigCheckBox(
+            parent=self._subcontainer,
+            position=(50, v),
+            size=(self._sub_width - 100, 30),
+            configkey='Show Ping',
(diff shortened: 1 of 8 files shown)


Loup-Garou911XD:
Really cool feature

efroemling:
This mostly looks good, but there's one thing I need to check on tomorrow before I pull it in:

SendReliableMessage() expects something with BA_MESSAGE_* as its first byte, so setting a BA_PACKET_SIMPLE_PING is incorrect and is probably erroring silently on the other end. We may need to define BA_MESSAGE_NULL or something for this purpose.

Basically BA_PACKET_* is the header byte for low level raw udp packets the app sends out. One type of packet is a 'GamePacket' which contains compressed commands for a game and also its own BA_GAMEPACKET_* header byte.  And lastly there is a reliable/unreliable messaging system built on top of GamePacket; each of these messages has a BA_MESSAGE_* header byte. Hope that makes sense. Check out networking.h for more info.

efroemling:
Ok I'll pull this in now and dig into the BA_MESSAGE_* part. I'm guessing I'll just add a BA_MESSAGE_NULL type message. As long as older versions of the game gracefully ignore those we should be good.
Thanks again for putting this together!

PR #537, opened 2022-10-24
Update _gameactivity.py

fraq-sss:
## Steps

## Type of Changes

### DIFF
--- a/assets/src/ba_data/python/ba/_gameactivity.py
+++ b/assets/src/ba_data/python/ba/_gameactivity.py
@@ -105,7 +105,7 @@ class GameActivity(Activity[PlayerType, TeamType]):
 
         This default implementation simply returns the 'name' class attr.
         """
-        return cls.name if cls.name is not None else 'Untitled Game'
+        return cls.name or 'Untitled Game'
 
     @classmethod
     def get_display_string(cls, settings: dict | None = None) -> ba.Lstr:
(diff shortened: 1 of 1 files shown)


assets/src/ba_data/python/ba/_gameactivity.py
```diff
@@ -959,7 +959,7 @@ def spawn_player_spaz(
         # Move to the stand position and add a flash of light.
         spaz.handlemessage(
             StandMessage(
-                position, angle if angle is not None else random.uniform(0, 360)
+                position, angle or random.uniform(0, 360)
```
Dliwk, line 962: In case `angle` is 0 that won't work as expected.
```py
>>> 0 or 1337
1337
```

Dliwk: But 'Optional or Default' syntax looks good in general. Note that it simply evaluates to `x if bool(x) else y` https://stackoverflow.com/a/3914686/17444908

Dliwk (changes requested):

efroemling:
Thanks for the contribution, but I think I'll pass on this one.

It does certainly make the lines look cleaner, but as @Dliwk mentioned above there can be unexpected corner cases where a  user might want to explicitly pass a value that evaluates to False. I've been burned a few times doing that sort of thing myself. For instance, if we set up something like:
`starting_powerups = passed_powerups or ['sword', 'club', 'rockets']`
..it looks reasonable but we've now made it impossible to start with an empty set of powerups.

So these days whenever I'm treating None as a special case I try to explicitly do `foo is None` instead of a bool evaluation even if that means the code is less pretty. Under the hood, 'is' simply compares the two objects' pointers so it shouldn't be too inefficient.

PR #538, opened 2022-10-24
Default slow motion init

fraq-sss:
## Steps

## Type of Changes


Freaku17:
Why do you even exist 😂😂

efroemling:
Thanks for the contribution, but for now I'd like to avoid adding common things like this that don't necessarily apply to all game types (even if they apply to significant numbers of them).

My hope for the future is to come up with a mechanism where something like a 'GameModifier' can be defined which can be applied to any game type; then it will be easier to add things like Epic Mode or Flying Mode or other new modes without having to add all those options to each game type.

PR #545, opened 2022-11-05
Plugin Settings

vishal332008:
## Description
A Small Plugin Settings in Plugin Window.
It can enable all plugins and disable all plugins in one go.
And there is an option for auto-enabling new plugins.

You have to add `"Auto Enable New Plugins": true,` in config.json for this to work.

And I do know that I haven't used `ba.Lstr` function for few texts in pluginsettings.py.
I will soon add another commit to fix it.

image

image


vishal332008:
We should add 'pluginSettingsText', 'pluginsEnableAllText', 'pluginsDisableAllText' and 'AutoEnableNewPluginsText' in languages.

efroemling:
Looks great to me; thanks for doing this!
I'll go ahead and add translation entries for the stuff here.
One note:
doing `ba.app.config['Auto Enable New Plugins']` will error for most people since they won't have that value in their config dict.
I'll change it to `ba.app.config.get('Auto Enable New Plugins', True)` which is safer; it'll default to True in that case.

efroemling:
Another similar note:
Instead of `plugs = cfg['Plugins']`
its best to do `plugs = cfg.setdefault('Plugins', {})` - this will do the right thing if that dict doesn't exist in the config.
(I'm adding that on my end; just wanted to mention it).

PR #549, opened 2022-11-27
Commit Last Manual Party Connect Port to config

ritiek:

## Description

Noticed when connecting to a friends LAN party through manual IP and when the server's port is anything other than 43210, it would still show up the last connection port on the client game as 43210 which doesn't seem right.

## Type of Changes


Dliwk (approved):

efroemling:
Thanks!!

PR #553, opened 2022-12-21
Spaz.curse(): self.curse_explode Timer Behavior Change

EraOSBeta:
## Steps

- [x] Add videos showcasing before and after the patch

## Description
with the current behavior, if you get cursed, then get a med-pack, and then get cursed again all within 5 seconds of getting the first curse powerup, you will explode based on the timer set from the first curse, meaning that you can get the second curse and explode 1 second after getting it instead of 5 (never seen it happen in vanilla bs but it seems possible), this small 1 line patch fixes that by making the timer behavior similar to other powerups

Old Behavior:

https://user-images.githubusercontent.com/100019405/208839409-efa1ab6d-d8a3-4a39-b310-06530a2f505d.mp4

New Behavior:

https://user-images.githubusercontent.com/100019405/208840095-f5b48347-b94f-4043-8a7b-7e3f692bc3f8.mp4

## Type of Changes


komasio71:
👏

Dliwk (approved):
I'd set `self._curse_timer` to None when handling health powerup message, but it seem to be unnecessary. :man_shrugging:

EraOSBeta:
> I'd set `self._curse_timer` to None when handling health powerup message, but it seem to be unnecessary. 🤷‍♂️

Yeah, curse_explode checks to see if you're cursed or not so it doesn't really change anything, also can you run the checks on latest commit again? Black committed fail on the last one 🥲

efroemling:
Good catch; thanks!

PR #557, opened 2023-01-09
Update spaz.py

ElSBP:
## Steps

## Description
Grammar tidying.

### DIFF
--- a/assets/src/ba_data/python/bastd/actor/spaz.py
+++ b/assets/src/ba_data/python/bastd/actor/spaz.py
@@ -376,7 +376,7 @@ class Spaz(ba.Actor):
         Called to 'press jump' on this spaz;
         used by player or AI connections.
         """
-        if not self.node or self.frozen or self.node.knockout > 0.0:
+        if not self.node:
             return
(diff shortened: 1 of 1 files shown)


efroemling:
Hi,
Thanks for the contribution!

However, the C++ layer already checks for frozen/knockout states when determining the effects of button presses so I think this is technically redundant. Setting the 'pickup_pressed'/'jump_pressed' attrs is simply how we communicate the raw button states down to the C++ layer where that logic happens. So I don't think adding extra logic here is ideal. Also Ideally the rate limiting and cooldown logic that are there should happen in the C++ layer too for that same reason (but there are backwards-compatibility reasons why they are currently in the Python layer).

Anyway, If you'd like to submit just the grammar tidying portions of this I'd be happy to apply it..

efroemling:
> Hi, i think this is not working as it should on pc, i found that when i spam the jump key while i'm frozen, its like its jumping without being frozen, i check this on android but i think the error is only on pc .

Interesting.. maybe some of the logic in spaz_node.cc is not properly ignoring frozen/knockout state if you're seeing that? Maybe we can fix it there at the source?..

PR #558, opened 2023-01-18
Update spaz_node.cc

vinnytherabbit:


efroemling:
Cool; this looks like it probably fixes things. Thanks for looking into that. Though I suggest we sit on this for just a bit until 1.8's scene-versions are ready so we can apply the change without having to jump through a bunch of backwards-compatibility hoops.  (see an explanation in this other PR -> https://github.com/efroemling/ballistica/pull/530#issuecomment-1386402607). I should be working on scene-versions soon; after I finish a few last bits of polishing on 1.7.x stuff.

efroemling:
> Sounds good, if you want you can merge the previous pr, now it only has grammar fixes.
> #557

Great; thanks! Will do...

PR #559, opened 2023-01-18
filtering unavailable maps from playlist

imayushsaini:
## Description
Filter out unavailable maps from a playlist to avoid blank screen/crashes.

Right now if playlist contains a map which is not installed on server then server stucks on scorescreen.

## Type of Changes


efroemling:
Good fix; thanks! I will probably tweak this slightly and add a ba.MapNotFoundError instead of raising an ImportError (which makes it sound like the problem is related to an import failing) but the logic looks sound to me.

PR #560, opened 2023-01-18
ffa scoring

Dliwk:
## Steps

## Description
Currently if multiple players receive same maximum number of points, they both get +8. This also annoying when there is only one guy who e.g. was the only one who held flag for all 30 seconds (others 0) and this guy will receive +8 meanwhile other receive +4. This PR makes it that player with same number of points receive that much series points as would receive last of them (not first, as now).

|name|final before|final now|game|
|--|--|--|--|
|spaz|+8| +8|100|
|zoe|+4| +2|50|
|kronk|+4| +2|50|
|pascal|+2| +1|0|

## Type of Changes

### DIFF
--- a/assets/src/ba_data/python/ba/_freeforallsession.py
+++ b/assets/src/ba_data/python/ba/_freeforallsession.py
@@ -74,9 +74,15 @@ class FreeForAllSession(MultiTeamSession):
             # Award different point amounts based on number of players.
             point_awards = self.get_ffa_point_awards()
 
-            for i, winner in enumerate(winners):
+            award_index = -1
+            for winner in winners:
+                award_index += len(winner.teams)
                 for team in winner.teams:
-                    points = point_awards[i] if i in point_awards else 0
+                    points = (
+                        point_awards[award_index]
+                        if award_index in point_awards
+                        else 0
+                    )
                     team.customdata['previous_score'] = team.customdata['score']
                     team.customdata['score'] += points
 
--- a/assets/src/ba_data/python/bastd/activity/freeforallvictory.py
+++ b/assets/src/ba_data/python/bastd/activity/freeforallvictory.py
@@ -53,7 +53,6 @@ class FreeForAllVictoryScoreScreenActivity(MultiTeamScoreScreenActivity):
             reverse=True,
             key=lambda p: (
                 p.team.sessionteam.customdata['score'],
-                p.team.sessionteam.customdata['score'],
                 p.getname(full=True),
             ),
         )


efroemling:
This might be a good excuse to start a discussion in general on how people feel FFA scoring should work.

- On one hand I can appreciate that it would be frustrating if one player gets a 'flawless victory' in a minigame which gives them 8 points but then all other players who tied with 0 each get 4 points for doing nothing.
- On the other hand to me it feels unintuitive dropping scores to be equal to the minimum of the tying group. It would be like someone winning a race and getting a 1st place ribbon and the next 2 people tying and each getting a 3rd place ribbon. I can see the motivation but it feels unusual to me. But maybe others disagree?

It makes me wonder if score distribution should be user-selectable or up to the individual mini-game; perhaps having just a few options such a classic (1st, 2nd, 3rd, 4th tier scores) and maybe a proportional option where scores get distributed based on mini game scores (so the aforementioned 'flawless victory' person would claim all points for the round).

Thoughts?

efroemling:
> Apparently, I've started a discussion about this some time ago #546.

Oh hmm maybe we should move the discussion there then; but I'm lazy at the moment so I'll just reply here.

> I imagine improving playlist editor, so it also allows configure entire session, like scoring system, points-to-win, etc.

Yes I think this would be great and would be happy to accept any such changes. Though the playlist editor code in general is rather old and janky and I feel like it could probably use a modern more clean/flexible rewrite.

On that topic I should mention that I hope to move away from 'teams' and 'free-for-all' sessions/playlists being such officially defined static things. I'd love for those concepts to move to bastd and for it to be possible to come up with other completely different ways to run a series of games - stuff like multi-team mode, or free-for-all mode where players get eliminated each round, or modes with more advanced settings/etc. Individual games shouldn't need to have knowledge of those details and should be able to be plugged in anywhere.

> And I'd like to have an interface like game settings, but with more general things, defined not by game itself. Something like that you can say "let's use mod1, mod2, and mod4 for this game in this playlist". But we need to come up with something for mods which override stuff to do their thing. Ideally if mods would not be required to override anything and only use provided interface, but we can also think about something like
> 
> ```python
> class MyPerGamePlugin(ba.Plugin):
>     def apply_for_game(game: type[GenericGameType]) -> type[GenericGameType]:
>         class ModifiedGame(game):
>             ...
>         return ModifiedGame
> ```
> 
> And session will apply all selected plugins when instantiating a game.

Yes I completely agree and have been thinking about this idea for a while. I'd love for things like epic-mode to be able to be defined as such a per-game-plugin for example. Or new powerup types. Or replacing TNT with some other super-powerup. Or replacing all spazzes with some other actor type. Or whatever else.

I would just hope we can do it in a way where it's not confusing to players. For example, epic mode currently adds 'Epic' to the game name so people know to expect slow motion. So maybe the list of plugin names/icons could show up when the game name is announced or something so people could learn what to expect.

I'm taking some time today and starting work on 'AppComponents' which will let plugins easily override parts of app behavior; maybe we can start learning from that what might work well to allow overriding things within games.

Also as cloud stuff starts coming online I'm planning on doing a bit of a 'clean room' implementation of things like Actors and GameActivities where we can experiment with stuff like this without having to touch the giant web of existing mini games and Actors and whatnot bundled with the game. Since a cloud server match will only need to run a single mini game and then deliver its results back to the master-server, we can go crazy with changes without worrying about compatibility. And then maybe when we come up with a good system we can back port everything back to the static bundled stuff in the game so it works offline/etc.

> > On the other hand to me it feels unintuitive dropping scores to be equal to the minimum of the tying group. It would be like someone winning a race and getting a 1st place ribbon and the next 2 people tying and each getting a 3rd place ribbon. I can see the motivation but it feels unusual to me. But maybe others disagree?
> 
> What about distributing points for remaining players based on how close to the finish they are?
> 
> > maybe a proportional option where scores get distributed based on mini game scores
> 
> Interesting. I only think that mini game scores should be mapped to series scores somehow like [minimum; maximum] -> [0; 8], so people in, for example, the same aforementioned race, who finished in 145s and 131s (and one didn't finish), will get 8 and 4 points instead of 8 and 7, respectively.
> 
> Many scoring mechanics can be imagined, so it may be good to leave them all to users and let them choose whatever they want.

There's definitely lots of factors to consider and I think positives and negatives to each option.
For instance, if you award points in a race based on how close to the finish you are:
- positive: more 'fair' in that performance maps more directly to score
- negative: less exciting in that there's less at stake - barely losing a race doesn't really matter much since instead of 8 points for first you'll simply get 7 (or maybe still 8 depending on the math)

I think stakes are a big part of why sports finals are so exciting - one team comes away with all the glory and the other comes away with nothing but sadness even though they might have played just as well throughout the whole season to get that far.

Maybe, as you mention, the right thing to do is just give people the option to override score tallying per playlist and they can decide. I could see the number of players and the selection of mini games having an effect on what sort of scoring would work best too. For instance, a super simple scoring mechanism where only the 1st place player gets 1 point could be exciting and feel less 'fuzzy' but it would probably only work well with 3 or 4 players and not 7 or 8 (since the 'bad' players would feel like they have no chance at any points then).

Anyway, I'd certainly be open to adding different score tallying as options to the existing FFA playlists. I guess questions are still outstanding as to whether individual games should be able to influence this as well or whether that would be too confusing.

Dliwk:
No free time in the foreseeable future, so I'll close it for now. Maybe one day...

PR #564, opened 2023-01-23
Adding Confirm Window

vishal332008:
Just a little update

### DIFF
--- a/assets/src/ba_data/python/bastd/ui/settings/pluginsettings.py
+++ b/assets/src/ba_data/python/bastd/ui/settings/pluginsettings.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 import ba
+from bastd.ui.confirm import ConfirmWindow
 
 if TYPE_CHECKING:
     pass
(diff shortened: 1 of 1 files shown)


efroemling:
My gut feeling is that a confirm window is overkill for that functionality. I feel like a confirm window would be appropriate for something more extreme like *deleting* all plugins. That being said, I'm happy to include pull it in if you disagree.

vishal332008:
Many people have a lot of plugins installed in their game. When they click the enable all / disable all button, it just enables/disables all the plugins and if they want to switch to their old manually selected plugins, they have to scroll down the whole plugins screen which has a lot of plugins. So I thought that a warning/confirmation window would be useful. If you feel like it is not needed, no problem in that. 👍

efroemling:
> Many people have a lot of plugins installed in their game. When they click the enable all / disable all button, it just enables/disables all the plugins and if they want to switch to their old manually selected plugins, they have to scroll down the whole plugins screen which has a lot of plugins. So I thought that a warning/confirmation window would be useful. If you feel like it is not needed, no problem in that. 👍

Ok; sounds reasonable. 👍 Thanks once again!

PR #570, opened 2023-02-05
We can't Wave arms if we're knocked out

vinnytherabbit:


Dliwk:
Just mentioning here current behavior:

spazwavinghishands

PR #584, opened 2023-04-16
Fixed the CTF Flag Bug

SoK05:
## Steps

## Description
There was an issue on how the flag return was handled. When the player who at the time was returning the flag died, the counter wouldn't stop and the flag would still return to the base.  Funny part? This bug would persist through the entire match. The only way to temporarily fix it was to throw the flag out of bounds.

I provided a simple fix, so now when the player who returns the flag dies, the flag won't return by itself.

Here's a quick presentation on how the flag behaves now:
https://youtu.be/ItF1f7mibDs

## Type of Changes


efroemling:
Sorry for letting this sit for a while!
I'm gonna close this since it applies to the pre-1.7.20 layout but I'll apply the change manually. Thanks for the fix!

PR #590, opened 2023-06-09
Categories in Plugins Window

vishal332008:
Same PR , 3rd time.... Continues #588 .
Confirmed working.


efroemling:
Thanks for adding this, and sorry for the huge delay pulling it in 😬

PR #591, opened 2023-06-10
Small Bug in Plugins Window

vishal332008:
This shows the gray color for disabled plugins.


efroemling:
Thanks for the fix! Sorry if I broke something with my code tidying.

Issue #600, opened 2023-07-09
BallisticaKitHeadless.exe does not find python


TheNotEasy:
### Description
When the BallisticaKitHeadless opens in "windows_x86_server/release/dist" it works. But if we run BallisticaKitHeadless outside of the directive, error appears

### Steps to reproduce
1. Launch executable outside
2. Bug!

### Machine
**Platform**: Windows 11 /  Ubuntu 22.04.2 LTS
**BombSquad version**: [1.7.21](https://github.com/efroemling/ballistica/releases/tag/v1.7.21)

### Extra
Console logs:
```
Python path configuration:
  PYTHONHOME = (not set)
  PYTHONPATH = (not set)
  program name = 'python'
  isolated = 1
  environment = 0
  user site = 0
  safe_path = 1
  import site = 0
  is in build tree = 0
  stdlib dir = 'D:\ballistica\build\prefab\full\windows_x86_server\release\dist\Lib'
  sys._base_executable = 'D:\\ballistica\\build\\prefab\\full\\windows_x86_server\\release\\dist\\BallisticaKitHeadless.exe'
  sys.base_prefix = 'D:\\ballistica\\build\\prefab\\full\\windows_x86_server\\release\\dist'
  sys.base_exec_prefix = 'D:\\ballistica\\build\\prefab\\full\\windows_x86_server\\release\\dist'
  sys.platlibdir = 'DLLs'
  sys.executable = 'D:\\ballistica\\build\\prefab\\full\\windows_x86_server\\release\\dist\\BallisticaKitHeadless.exe'
  sys.prefix = 'D:\\ballistica\\build\\prefab\\full\\windows_x86_server\\release\\dist'
  sys.exec_prefix = 'D:\\ballistica\\build\\prefab\\full\\windows_x86_server\\release\\dist'
  sys.path = [
    'lib',
    'DLLs',
  ]
FATAL ERROR: Precondition failed @ core_python.cc:115: !PyStatus_Exception(status)
CPP-STACK-TRACE-BEGIN:
0   (unknown symbol name), address 0xFFCFCC78.
1   (unknown symbol name), address 0xFFCFCC78.
2   (unknown symbol name), address 0xFFCFCC78.
3   (unknown symbol name), address 0xFFCFCC78.
4   (unknown symbol name), address 0xFFCFCC78.
5   (unknown symbol name), address 0xFFCFCC78.
6   (unknown symbol name), address 0xFFCFCC78.
[...]

PR #601, opened 2023-07-09
Update code.py

TheNotEasy:
Switching from calling black with subprocess to direct use

## Description
Transition from using black via subprocess to calling black format function directly to avoid path limit exceeded error on windows and (i think) speed up format process

## Type of Changes

|   | :hammer: Refactoring   |
|    | :scroll: Docs          |

### DIFF
--- a/tools/efrotools/code.py
+++ b/tools/efrotools/code.py
@@ -312,7 +312,7 @@ def format_project_python_files(projroot: Path, full: bool) -> None:
 def format_python_str(code: str) -> str:
     """Run our Python formatting on the provided inline code."""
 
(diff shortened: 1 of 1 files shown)


TheNotEasy:
oops, i forget about checking ci, coming soon...

TheNotEasy:
I cannot test make preflight cuz of my [issue](#600), but I can tell that it will fail cuz black is not included in "lib" file

TheNotEasy:
I cannot test make preflight cuz of my [issue](#600), but I can tell that it will fail cuz black is not included in "lib" folder

PR #603, opened 2023-07-12
Fix possible DoS vulnerability

Dliwk:
## Description
A lot of nested multipart messages may trigger an allocation of a really huge amount of memory, which OS couldn't handle.

This PR disallows nested multipart messages.

## Type of Changes


efroemling:
Thanks for the fix!

PR #612, opened 2023-07-30
Made the v2 signin link copyable

vishal332008:
I thought of making a copy button but this was more efficient I think...
Btw, it looks good on mobile. If it is not aligned correctly in tab or laptop, plz comment.

Edit : works good on all ui scales


vishal332008:
The last commit is a small fix.
In profile edit window, if you add spaces before your name in profile name and it exceeds 10 characters, in the right, it shows that name in game is displayed with blank spaces in the start. But when saving, game removes the blank spaces. So the name shown by the text in the right side was wrong.

PR #616, opened 2023-08-30
Fixed PlayerSpaz using bs.apptime() instead of bs.time() for last_attacked_time

EraOSBeta:
## Steps

- [x] Update CONTRIBUTORS.md

## Description

PlayerSpaz should use the scene's time for handling things, but it accidentally got changed to use the app time instead during the transition to api 8

## Type of Changes

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -41,9 +41,8 @@
 - Modder
 
 ### Era0S
-- Fixed a single bug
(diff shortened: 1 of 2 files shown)


efroemling:
Thanks for the fix! I've run across a few other cases where I made that same dumb mistake (putting in app time where there should be scene time). Please holler if you come across any others.

PR #617, opened 2023-09-13
Fixed issue with hockey puck dying

EraOSBeta:
## Steps

## Description

There was always a bug annoying me when working with mods regarding how the hockey puck handles a non-immediate bs.DieMessage, causing the puck to respawn indefinitely, this PR ensures that this won't happen.

## Type of Changes


efroemling:
Good catch and thanks for the fix!

PR #618, opened 2023-09-22
Fixing a small error

vishal332008:
Fixing a small error...


efroemling:
Took me a while to see what real-world case this would fix, but I figured it out: if the user types spaces before or after their name, this properly reflects the fact that those will be stripped out for their final name. (let me know if there's another case I'm missing).

Thanks for the fix!

PR #624, opened 2023-09-30
A little update...

vishal332008:
I hope copying text directly is ok instead of copy button for the v2 account connecting link...

And a little fix.
Hoping it is ok.
(And i tested that code like before 1½ month so check it before you merge it)


efroemling:
Thanks for the fixes and the copy functionality!

PR #626, opened 2023-10-08
Few Changes

vishal332008:
Should add the NoServerFound text or if you had any other text matching it, you can add that. And... that's about it.

3 changes:
1. Changelog mistake.
2. Added cancel button in on-screen keyboard.
3. Added no server found text if searching for a server returns no servers.


efroemling:
Thanks for this! Sorry I can be slow pulling these in 😬

PR #627, opened 2023-10-14
Update _app.py

CrossJoy:
Solve error by removing the asterisk

## Steps

## Type of Changes

### DIFF
--- a/src/assets/ba_data/python/babase/_app.py
+++ b/src/assets/ba_data/python/babase/_app.py
@@ -807,7 +807,7 @@ class App:
                     _ = task_group.create_task(
                         self._run_shutdown_task(task_coro)
                     )
-        except* Exception:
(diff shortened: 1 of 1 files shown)


efroemling:
That isn't an error. Exception groups are a new thing in Python 3.11.
https://docs.python.org/3/reference/compound_stmts.html#except-star

Are you perhaps using an editor or something that doesn't understand those yet?

CrossJoy:
> That isn't an error. Exception groups are a new thing in Python 3.11. https://docs.python.org/3/reference/compound_stmts.html#except-star
> 
> Are you perhaps using an editor or something that doesn't understand those yet?

I should have known that 😅, My apologies. I just realised Im using an editor that uses 3.9 interpreter and not 3,11.

PR #628, opened 2023-10-24
Added the reject_recently_left_players plugin into the base game

EraOSBeta:
## Steps

- [x] Add Rikko to CONTRIBUTORS.

## Description
Imagine killing someone, them having to wait 7 seconds to respawn, and they just disconnect and reconnect and they're able to hop in the game immediately without having to wait even a second.

Originally a plugin by @rikkolovescats, this PR disallows any pb-id from joining the session for 10 seconds once the player leaves the session, this system works like the system in the c-layer that disallows people from joining for 10 seconds after they leave, but this system supports disconnections and reconnections, allowing a fair game.

It's also worth noting that this system does not stop people from joining if authenticate_clients is set to false in the config and the player sends a join request to the server before the server fetches the pb-id, this is intentional as the server has requested the client to not be authenticated.

## Type of Changes

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -43,6 +43,10 @@
 ### Era0S
 - Bug Fixer
 - Modder
+- Added a feature
 
 ### VinniTR
 - Fixes
+
+### Rikko
+- Created the original "reject_recently_left_players" plugin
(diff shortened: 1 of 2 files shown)


EraOSBeta:
Also Eric you add the CHANGELOG entry I'm too lazy.

efroemling:
Thanks for the improvement! My only comment would be to add some comments on what the code is doing, but I'll go ahead and do that.

PR #630, opened 2023-11-07
Update spaz.py

VinniTR:


EraOSBeta (changes requested):
Line 1160 should be reverted since it's supposed to tell the function and ultimately the activity who the source of the explosion is if it's not us

EraOSBeta (approved):

efroemling:
Thanks for the cleanup/fix!

PR #632, opened 2023-11-12
Time-pass

vishal332008:
You don't know how much suffering I went through to make this code. Don't ask me about why 6 commits for simple code. Ask that to mypy and pylint.

Ok... so pretty much 90% of the players play bombsquad online. So, every year they have to update their game just to change the year. So I thought that adding the latest year when new year comes will be good.
And we don't have to be in a hurry to change the year in the code when new year comes.

### DIFF
--- a/src/assets/ba_data/python/bascenev1lib/mainmenu.py
+++ b/src/assets/ba_data/python/bascenev1lib/mainmenu.py
@@ -83,7 +83,7 @@ class MainMenuActivity(bs.Activity[bs.Player, bs.Team]):
                         'scale': scale,
                         'position': (0, 10),
                         'vr_depth': -10,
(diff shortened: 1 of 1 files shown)


EraOSBeta (commented):
I don't think this change should be made for 2 reasons:
- There really isn't a rush to update the footer anyways.
- A copyright footer is supposed to be updated every year if there has been contributions made to it in that year, not automatically.

I think a better approach would be to have a program check or automatically update the footer every time a commit is pushed to the repository, this could be CI failing if it finds an outdated footer on push or `make update` automatically updating copyright footers if found.

vishal332008:
Nice, now I will go ahead and close this. Atleast I learnt how CI can mess with your code.
Btw, this was just a side code. I was working on smth else. Thought I can time pass with this. 😐

PR #633, opened 2023-11-13
A small change

vishal332008:
1. Fixed a screen message that no one will ever see.
2. Displaying number of plugins added to the game.
3. Displaying 'No Plugins Installed' if no plugins are there.
4. Displaying the old messages as soon as you press 'Unmute Chat'.
5. Added 'Add to Favorites' button in chat window for adding server to Favorites List.
6. Displaying 'No Parties Added' if no parties are there in Favorites Tab.
7. Showing the Characters of the specific names in Profiles Window.
8. Added "Random Name" button (used replay icon) which gives random name to the player.
9. Fixed a bug where no server is selected default in favourites tab.
10. Fixed a bug where no replay is selected default in watch tab.
11. Fixed a bug where no profile is selected default in profile tab.

A thing about point 2, I am displaying only the number and not text along with it because.. well because people have common sense. As soon as the player adds a plugin, the player will understand. So yeah.

And can you add seperate change logs for each point? If it is fine for you Eric.

Screenshot_20231122_111255
Screenshot_20231122_110737
Screenshot_20231126_134548
Screenshot_20231122_110631
Screenshot_20231123_180443


EraOSBeta (approved):

EraOSBeta (changes requested):
Point 5 should be reverted and held back until Eric adds an actual function for getting full info about the party we're in or Smoothy's hacky way of finding the current party's info should be implemented instead.

vishal332008:
It is okay to keep point 5 here itself. It is fine for me if Eric takes his time to pull this. And if Eric checks this PR, it will only take him like inside 1 hour to fix the problem of point 5. So no problem.

EraOSBeta (approved):

EraOSBeta (changes requested):
Add to favorites button should be removed when we're the host (if this is possible right now).

EraOSBeta (changes requested):
The profile browser should load Spaz's icon if a profile has an invalid character, it currently just throws an error and doesn't show any profiles if any of them have an invalid character.

vishal332008:
> Add to favorites button should be removed when we're the host (if this is possible right now).

We didn't quite discuss the extent of add to favorites. I already said in the main comment (I edited it a few days before, so not your mistake if you didn't read it) that until Eric decides what he will add, it is waste of adding stuff which can be replaced. After Eric checks and says something about adding internal function or smth else, it's best to leave it like that.


> The profile browser should load a fallback icon if a profile has an invalid character, it currently just throws an error and doesn't show any profiles if any of them have an invalid character.

Ok. Will add that.

EraOSBeta (approved):

efroemling:
Some very nice additions; thanks!

Thoughts:
- For #5, I'd rather add a function to get the name/address of the current server. Could we take that one out for now and do it in a separate PR?
- @EraOSBeta mentioned the profile-browser doesn't handle missing icons; could you fix that?
- I think maybe just saying 'random' for the profile name button would work better; right now it looks a bit like it refreshes something.

Anyway, if you can make those few tweaks, I'd be happy to add whatever translation entries are needed and pull this in.

vishal332008:
Done

EraOSBeta:
I think the title of this PR is no longer correctly representing this PR

efroemling:
Ok I added bascenev1.get_connection_to_host_info_2() which will provide server name, address, and port (when connected). Let me know when you've got that wired up and I can pull this in 👍

vishal332008:
Can you release builds so I can actually test it?

efroemling:
Ok new test builds are up.

EraOSBeta:
Note: you can use [CD](https://github.com/efroemling/ballistica/actions/workflows/cd.yml) for obtaining the newest test builds (excluding android builds) from now on

vishal332008:
Ok. I finished the code of favourites button. Found a bug on the way. I fixed it in this PR's last commit. Now you can pull this one in so I can create new PR.

vishal332008:
Or should I add the favourites button code in this PR itself?

efroemling:
> Or should I add the favourites button code in this PR itself?

Up to you. I'm ready to pull this PR in anytime so let me know if you'd prefer to add that here first or in a separate PR.

vishal332008:
I added it here itself. You can pull this PR now.

efroemling:
Finally pulling this in. Thanks for all the improvements!
Do you by chance have a list of all the text resource strings I'll need to add?

efroemling:
Actually I took a pass through and added the resource strings I saw. Just let me know if I missed anything.

PR #642, opened 2023-12-24
A Button for Pausing Replay!

vishal332008:
## Description
So.. description... it pauses and resumes replays with a click of a button.
And.. have to add pauseIcon ( |> )and ResumeIcon ( | | ) textures.

## Type of Changes
| ✓   | :sparkles: New feature 

## Video
https://github.com/efroemling/ballistica/assets/76160371/47ea004c-a1bd-40c0-b9e4-0772eee2be15

(This is a version where we used text instead of texture.)


EraOSBeta:
## Preflight
This PR is ready to be merged is it succeeds most tests using a local build instead of a prefab (which is why CI fails), the only changes that need to be made after merging is adding a newline at the end of all the edited files

vishal332008:
Added the new lines. So only problem for Eric will be to make new prefabs which contain these new changes.

EraOSBeta:
## Preflight
All checks passing :+1:

efroemling:
Thanks for this! It looks like CI is failing because some of the checks are using the prebuilt binaries instead of building them freshly which would pick up your changes to the native layer. Maybe it would be worth switching to building everything, though that would slow CI down quite a bit.

EraOSBeta:
> Thanks for this! It looks like CI is failing because some of the checks are using the prebuilt binaries instead of building them freshly which would pick up your changes to the native layer. Maybe it would be worth switching to building everything, though that would slow CI down quite a bit.

Trust me I tried, it just won't work, CI just complains about SDL2 or something when you force it to build it's own thing for running checks.

efroemling:
> > Thanks for this! It looks like CI is failing because some of the checks are using the prebuilt binaries instead of building them freshly which would pick up your changes to the native layer. Maybe it would be worth switching to building everything, though that would slow CI down quite a bit.
> 
> Trust me I tried, it just won't work, CI just complains about SDL2 or something when you force it to build it's own thing for running checks.

That was more of a rhetorical comment to myself :). It definitely would be possible to get CI checks to build a binary first (some of the other CI checks already do build binaries) but it would take a bit of tinkering and I'm not sure if the extra few minutes per check would always be worthwhile. Something to think about though.

PR #647, opened 2024-01-01
Update README.md

komasio71:
just correcting...


vishal332008:
The Xbox change is okay, but change back toaster to capital t. It's kinda funny. People will think Toaster is also a platform like Xbox and PlayStation.

efroemling:
Thanks for the fix!

PR #651, opened 2024-01-10
Reset Mapping button in controllers config menu

3alTemp:
## Description
[As suggested by @komasio71](https://discord.com/channels/1001896771347304639/1175427047124897812/1175427047124897812), this implements a "reset" button in the "Configure ``X``" menus that sets your button mappings to their default values. That's it.

https://github.com/efroemling/ballistica/assets/79161340/d5613adf-5777-4bd2-8af6-89479d074aaa

image
image
image

I don't think this requires much explanation; it assists you on wiping your button binds in case you set them to something silly and non-functional, if you're not happy with your current binds, or literally any other reason on which you would need to clear them out. Not a big change but it would be appreciated by tons of people and controllers alike :).

# EFRO NOTE:
## Now, look at this very cool table telling you this is a new feature:


komasio71:
my idea!! :D (this comment was supposed to be posted a week now but anyway)

3alTemp:
I'm too tired to change the whole input reset ordeal into another file, efro please just add your magical ``# pylint: disable=too-many-lines`` if you agree...

efroemling:
This is a great addition; thanks! My one bit of feedback would be that I think the button should be a bit less prominent. I'd prefer to have 'Cancel' and 'Save' be the primary two actions users see. Could the reset button be a more muted color, smaller, and maybe tucked into the bottom or corner? Perhaps similar to the 'advanced' button in the controller window?..

3alTemp:
> This is a great addition; thanks! My one bit of feedback would be that I think the button should be a bit less prominent. I'd prefer to have 'Cancel' and 'Save' be the primary two actions users see. Could the reset button be a more muted color, smaller, and maybe tucked into the bottom or corner? Perhaps similar to the 'advanced' button in the controller window?..
Sure! I'll move the button to a corner real quick! (not as quick as I'll need to make sure it doesn't break on smaller UI sizes 🤠)

efroemling:
Alternately I could see this also working as a menu entry under a '...' button. That would also give us a place to add any other functionality later (and the advanced button could go there too). Just a thought.

3alTemp:
True, I'll follow to that and move them both in there.

3alTemp:
Oh well, I'll leave the CI thing there; I've been trying to install it on VSC for a while, but it keeps plaguing the project with nonsensical ``Mypy(no-any-unimported)`` errors.

efroemling:
Thanks for this!

PR #652, opened 2024-01-11
Bug Fix and Passing CI

vishal332008:
## Description
Sometimes modders/players use plugins to enable pro. In coop, if you keep hard mode and disable the plugin which unlocks pro, you will see that the game is still in hard mode and you can play all the hard mode levels even when lock sign is shown. This PR fixes this small annoying issue.

And fixing CI.

## Type of Changes


3alTemp (changes requested):
1. Is it really a bug if you have to use mods to trigger it?
2. You can do the same fix by simply changing [lines 103~105](https://github.com/efroemling/ballistica/pull/652/files#diff-de56926c292d16895f1ed02655562a56dda161765f5b71409a51700ad2af6e67R103-R105) to:
```py
        self._campaign_difficulty = (
            'easy'
            if not bui.app.classic.accounts.have_pro_options()
            else plus.get_v1_account_misc_val('campaignDifficulty', 'easy')
        )
```

vishal332008:
1. It is a bug if game behaves weird even after the mod is disabled and removed.
2. Yeah, I should. But I was trying to make the campaign difficulty to go back to default (easy)  instead of just changing it to easy everytime the code runs. Plus I was in a hurry, my mistake, shouldn't have created PR that soon.

3alTemp:
That is... debatable: a lot of mods can break game logic even after being uninstalled, see for example mods that install characters or maps to the game, removing the script itself will still leave code residue in the base game, and uninstalling doesn't guarantee the game will go back to 100% vanilla; even at the slightest discrepancy, tournaments would lock up.
Also, having the code set a value each time is run isn't inherently bad, but the good solution would then be setting the account difficulty value directly:
```py
        self._campaign_difficulty = plus.get_v1_account_misc_val(
            'campaignDifficulty', 'easy'
        )
        # Non-pro players shouldn't be able to have 
        # any mode other than easy selected.
        if (
            self._campaign_difficulty != 'easy'
            and not bui.app.classic.accounts.have_pro_options()
        ):
            self._campaign_difficulty = 'easy'
            plus.add_v1_account_transaction(
                {
                    'type': 'SET_MISC_VAL',
                    'name': 'campaignDifficulty',
                    'value': 'easy',
                }
            )
```

3alTemp:
Ah, you already made some changes. Epico 🤠

3alTemp (approved):
Yeah, these reformats are good.

efroemling:
This looks good to me; if you can resolve the conflicts I'd be happy to pull it in 👍

vishal332008:
> This looks good to me; if you can resolve the conflicts I'd be happy to pull it in 👍

Done

efroemling:
Schweeeeet. Thanks!

PR #654, opened 2024-01-12
Meteor Shower minigame end fix

3alTemp:
## Description
Meteor Shower has this really minor bug that "ends" the game twice if you die before the 5 second mark, usually resulting in the ringing sound effect playing twice and in vs teams mode setting the timer to 5.00s after the game finishes. This PR fixes that 🤠.

### Bug footage (sound on!)
https://github.com/efroemling/ballistica/assets/79161340/68173439-aa46-4a2c-89ef-0ae11b9db233

### Fix footage (sound on here too!)
https://github.com/efroemling/ballistica/assets/79161340/b8b824d1-d227-4757-819c-184a14f3cb3c

*A bit laggy because debug builds don't run that well on windo'*

## The table


efroemling:
Thanks for the fix! Ideally the game activity class should cleanly handle the case of being told the game is over multiple times, but that's not really possible with the current api so this gets the job one. 👍

PR #655, opened 2024-01-14
Update spazbot.py

VinniTR:
## Description
BomberBotLite always walks at the same speed like BrawlerBotLite, BomberBot doesn't do that.
ChargerBotPro should'nt have a shield.

## Type of Changes


efroemling:
I think it'd be preferable to keep BomberBotLite's behavior as it is since it's used in a few co-op levels. I'd be fine changing ChargerBotPro to remove the shield though since it's not used in any built-in coop stuff and it makes no sense when ChargerBotProShielded also exists. If you can take out the BomberBotLite bit I'd be happy to pull this in.

VinniTR:
now its ready!

efroemling:
Thanks for this!

PR #659, opened 2024-01-26
 Power-ups affecting melee combat

VinniTR:
## Description
Power-ups are affecting melee combat, why? Because these can be hit, you take knockback and this puts you at a disadvantage, It's like you punch a wall. 
In the case of hitting your opponent who was nearby, the damage will be less than that received by the great immovable box. Thanks to this pr, it will be less complicated to fight near the power-ups placed in the middle.
Also it doesn't make any sense that you can hit powerups and receive knockback if they disappear with touch, I think this makes it easier for you to grab them if you hit them

## Type of Changes

### DIFF
--- a/src/assets/ba_data/python/bascenev1lib/actor/spaz.py
+++ b/src/assets/ba_data/python/bascenev1lib/actor/spaz.py
@@ -13,7 +13,7 @@ from typing_extensions import override
 import bascenev1 as bs
 
 from bascenev1lib.actor.bomb import Bomb, Blast
-from bascenev1lib.actor.powerupbox import PowerupBoxFactory
+from bascenev1lib.actor.powerupbox import PowerupBoxFactory, PowerupBox
 from bascenev1lib.actor.spazfactory import SpazFactory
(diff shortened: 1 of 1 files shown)


VinniTR:
https://github.com/efroemling/ballistica/assets/71152012/b693ad40-22a7-4faa-a52a-130ab20127ec

VinniTR:
actual behavior

https://github.com/efroemling/ballistica/assets/71152012/88e48c88-767a-4200-b3bb-0fed73156e75

SoK05:
Oo this change seems cool! Gives a bit more depth - punch to get the powerup _just a bit_ faster. Neat!

efroemling:
Looks good to me; thanks! In principle I slightly don't like that this adds spaz behavior that is specific to the powerup class (ie: if someone wants to write a replacement powerup class it won't behave the same)... but pragmatically there's lots in spaz that is hard-coded in ways such as this, so we can worry about cleaning things up in scene-v2/etc.

PR #667, opened 2024-03-02
Support IPv6 address on manualtab

brostosjoined:
## Description
This is a simple fix that allows player to use ipv6 addresses on the manual tab to connect to servers and maybe their friends party. The method used earlier `socket.gethostbyname` only supports ipv4 so instead i switched it to `socket.getaddrinfo` that support both protocols.
Some video samples here of before and after.

https://github.com/efroemling/ballistica/assets/67740566/ea2ffe79-82b9-4fe1-8d3a-29723a899b37

https://github.com/efroemling/ballistica/assets/67740566/d3b5c066-adf9-45be-ac4f-30ba7e898c1f

## Type of Changes

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -50,3 +50,6 @@
 
 ### Rikko
 - Created the original "reject_recently_left_players" plugin
+
+### brostos
+- Added support for joining using ipv6 address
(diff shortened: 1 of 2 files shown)


efroemling:
Thanks for the fix!

PR #671, opened 2024-03-04
HEX support for advanced color picker

3alTemp:
## Description

Derivate from my work-in-progress BombSquad mod. I thought having support for HEX codes in the advanced color picker would be a neat, extra little feature that'd be appreciated by those people who look for more color accuracy (the current advanced color picker always rounds all values to 2 decimals).

https://github.com/efroemling/ballistica/assets/79161340/13365fcc-85e8-4b54-b44e-96a5a6a3e3a9

### NOTE:  For this to work properly, it requires the ``clayStroke`` texture. It was made using a UI window and adding some filters and masks, leaving it gray and texture enough to show the color like the game would.
clayStroke

## Oh no! Our table!


efroemling:
Nice addition!

I think my one nitpick is that it looks a bit broken with the 'clear text' X button overlapping the hex value. Could you widen everything just a bit or shrink that text box a bit or something so that doesn't happen?

3alTemp:
> I think my one nitpick is that it looks a bit broken with the 'clear text' X button overlapping the hex value. Could you widen everything just a bit or shrink that text box a bit or something so that doesn't happen?

That's a fair nitpick; I left it the way it is because the text looks nice centered and overall didn't look that crammed to me, but we could definitely make it space out the text like input queries usually do.

efroemling:
I should mention another fix would be to disable to 'clear text field' button for this widget. I think there's a flag to do that.

3alTemp:
I... did not know that was a thing! 😨

efroemling:
Oh looks like I'm misremembering. It doesn't draw that control in some situations such as when running on TV (since there's no way to press it there). But there's currently no way to explicitly deactivate it. I'll add one right now; should be very simple. Back in a few.

efroemling:
Ok just added an `allow_clear_button` arg to the `textwidget()` call if you'd like to use that. Holler if it doesn't work. 
7c45d16155fd03c138a30a07e3b1c6bde1344672

3alTemp:
It seems to be behaving properly 🤠

efroemling:
Thanks for this!

PR #679, opened 2024-03-15
Profiles icon tweak

3alTemp:
## Description
Profile character icons are rendered as ``buttonwidget``, yet they don't seem to do anything when clicked / tapped.
This commit changes them to ``imagewidget``, moves the profile text to the left and renders the images above them. The change is virtually unperceivable, but now clicking on icons acts as selecting the profile.

## Type'o Change


efroemling:
Thanks for the cleanup!

PR #689, opened 2024-04-09
having custom health amount on a spaz

### LINKED ISSUE #688: Alterable Health amount
I think we should let hitpoints (hitpoints_max) be changeable by default, ik someone else had done it before in their modpacks, but I want it to be changeable by anyone (who open internal code like me) easily, so they might add bots with extra health (what I did), special powerup (that could interfere with hitpoints) , or you name it.
`#code from spaz.py
    class Spaz(bs.Actor):
    """
    Base class for various Spazzes.

    Category: **Gameplay Classes**

    A Spaz is the standard little humanoid character in the game.
    It can be controlled by a player or by AI, and can have
    various different appearances.  The name 'Spaz' is not to be
    confused with the 'Spaz' character in the game, which is just
    one of the skins available for instances of this class.
    """

    # pylint: disable=too-many-public-methods
    # pylint: disable=too-many-locals

    node: bs.Node
    """The 'spaz' bs.Node."""

    points_mult = 1
    curse_time: float | None = 5.0
    default_bomb_count = 1
    default_bomb_type = 'normal'
    default_boxing_gloves = False
    default_shields = False
    hitpoints = 1000 #put hitpoint value here for it to be alterable on demand.
    hitpoints_max = 1000`

IMG_0329
_This is a bot I've custom made with extra health points._

#689

MW2917:
It's perfect for anyone who wanted an even tougher bot, powerups that can interfere with health (I already thought of steroids (Now to figure out how the hell do I make a powerup))  or any mods on mechanics they can think of without making this change manually everytime.

This closes #688 

## Type of Changes

| ✓   | :hammer: Fix what I think was a bug.   |
| ✓   | :scroll: Great for change modders         |

### DIFF
--- a/src/assets/ba_data/python/bascenev1lib/actor/spaz.py
+++ b/src/assets/ba_data/python/bascenev1lib/actor/spaz.py
@@ -68,6 +68,8 @@ class Spaz(bs.Actor):
     default_bomb_type = 'normal'
     default_boxing_gloves = False
     default_shields = False
+    hitpoints = 1000
+    hitpoints_max = 1000
 
     def __init__(
         self,
@@ -174,8 +176,6 @@ class Spaz(bs.Actor):
                     setattr(node, attr, val)
 
             bs.timer(1.0, bs.Call(_safesetattr, self.node, 'invincible', False))
-        self.hitpoints = 1000
-        self.hitpoints_max = 1000
         self.shield_hitpoints: int | None = None
         self.shield_hitpoints_max = 650
         self.shield_decay_rate = 0
--- a/src/assets/ba_data/python/bascenev1lib/actor/spaz.py
+++ b/src/assets/ba_data/python/bascenev1lib/actor/spaz.py
@@ -69,7 +69,6 @@ class Spaz(bs.Actor):
     default_boxing_gloves = False
     default_shields = False
     default_hitpoints = 1000
-    default_hitpoints_max = 1000
 
     def __init__(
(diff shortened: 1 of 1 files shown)


EraOSBeta (changes requested):
It's not really useful since you can simply change those values the moment spaz is created; but if this goes through in my opinion it's better to do `default_hitpoints_max = 1000` instead of the 2 lines you added and then do `self.hitpoints = self.default_hitpoints_max` and `self.hitpoints_max = self.default_hitpoints_max` instead of the 2 lines you removed, since having to set hitpoints and hitpoints_max separately before spawning spaz doesn't make a whole lot of sense.

MW2917:
> It's not really useful since you can simply change those values the moment spaz is created; but if this goes through in my opinion it's better to do `default_hitpoints_max = 1000` instead of the 2 lines you added and then do `self.hitpoints = self.default_hitpoints_max` and `self.hitpoints_max = self.default_hitpoints_max` instead of the 2 lines you removed, since having to set hitpoints and hitpoints_max separately before spawning spaz doesn't make a whole lot of sense.

yeah just like the curse time thing. thx

EraOSBeta (approved):
Seems pretty good. Now we just need eric to approve workflow runs so we can see if all checks pass or not (I'm too lazy to run checks locally right now)

efroemling:
Looks good to me. Thanks for the improvement!

PR #691, opened 2024-04-22
If we're knocked out, we'll drop any object that we're holding

vinnytherabbit:
## Description
If we're knocked out and we're holding a object, we'll drop it immediately.

## Type of Changes


Loup-Garou911XD:
why do i feel i have seen this pr before?

EraOSBeta:
Doesn't this face the exact same issue as #530? since this stuff isn't server-side only

EraOSBeta:
Alright so I was on-board with having this be merged for scenev2 as one of the many minor engine issues that need fixing; but now I'm against having this in the C++ layer since mods could possibly use knocked-out spaz nodes that are holding other nodes as a way to create chain-like objects (maybe, haven't tested it). So I would personally recommend any logic of this kind (along with any other non-heavy-performance limitations currently in the spaz node code) to be moved to the Python layer.

PR #692, opened 2024-04-22
if we're knocked out, don't flail our legs and arms when airborn

vinnytherabbit:
## Description
Let's keep this here in the meantime.

## Type of Changes


EraOSBeta:
This can pass into scenev1 if flailing doesn't really affect anything in the physics simulation

efroemling:
This would *slightly* affect physics sims in some cases: older clients would be simulating with flailing limbs while the servers they are connected to would not, so there'd be some popping and resyncing until everyone got the update. But I don't think that would be too problematic.

Rather my main question with this would be: does it make things funnier or improve gameplay? The knockout mechanism is kinda meant to just mean that a character is too stunned to stand; I don't think it breaks any logic if they still wave their arms when flying through the air. Right now one can tell at a glance if a flying character is alive or dead by whether they are flailing, but this would complicate that. And I worry that a large percentage of events that launch a character through the air would also knock them out, meaning that we could essentially just be removing flailing, which is no fun.

So anyway I'm *probably* against this for those reasons, but I do want to be receptive to other opinions if you feel it is an improvement.

VinniTR:
@efroemling i think it's funny because the character would be like a rag doll every time he goes flying, whether due to a punch or an explosion, the death animation could be modified a little to avoid confusion with living and dead people, if that's what it is what you mean...

EraOSBeta:
@VinniTR can you show us a comparison video between the old behavior and the new one so we could compare the funni value?

VinniTR:
> @VinniTR can you show us a comparison video between the old behavior and the new one so we could compare the funni value?

i dont know how :cc

vishal332008:
@efroemling @VinniTR @EraOSBeta 
Normal:
https://github.com/user-attachments/assets/53f08fc0-ad99-444d-8255-37c4a7775a4d

New:
https://github.com/user-attachments/assets/e277f614-b44d-456d-9d4a-a221b6f4b140

EraOSBeta:
Seems to add a bit of funni, but I think it would be nice to have both this and the old behavior available in the game at the same time with the new behavior being a toggle in the advanced menu for a few versions; this way we could merge this for now while the people decide which one is better.

I think in general we need an "Experiments" section in the advanced settings to add all sorts of QoL changes there for stuff like this, when we aren't *really* sure if the change is better or worse.

MatiasMGz:
I think this subsection of "Experiments" should be in Settings->Advanced.

MatiasMGz:
I would like this to be added to the game anyways.

SlimeBois:
if theres toggle for that no flail ragdoll for modding, y ea

DinoWattz:
This disables flailing _only_ while knocked out right? This would make it easier to tell if your character is knocked out while airborne

> Right now one can tell at a glance if a flying character is alive or dead by whether they are flailing, but this would complicate that.

There's a bunch of indicators already though, like..

- The character won't be screaming while falling if the same is dead
(_**but I think it'd look odd if they screamed while knocked out without flailing**_)
- The character will either be glowing bright red or be darkened
- The character's health bar will be flashing red and if it has a name, the name will be desaturated

but we still need to test if those indicators would be good enough by themselves

> And I worry that a large percentage of events that launch a character through the air would also knock them out, meaning that we could essentially just be removing flailing, which is no fun.

Bomb explosions also knocks out characters if I remember correctly, I CAN'T imagine Spaz not flailing (_**nor screaming**_) after being launched by an explosion.
_Would there have to be two type of knockouts? Will it need to be damage type based? Does it need to be moddable?_

Geez this really is complicated.

PR #699, opened 2024-05-18
Docker build improvements

Loup-Garou911XD:
## Description
- Removed unnecessary packages getting installed(image size reduced to 216 MB :)) 
- Using -f flag for `docker build` instead of copying Dockerfile to project root
- Added github action cd target for making docker images
- Added `docker-clean` and `docker-save` make targets

## Type of Changes


efroemling:
Thanks for this!

Loup-Garou911XD:
@efroemling https://github.com/efroemling/ballistica/pull/699/files#diff-23563e327b2d88cb3e33ae954f1b7febca76cac5899198d0938854eab302c072R70
here we are generating random machine-id, but that is done when building the image
so if we share the docker image
it will have same machine-id
although through testing i got to know that the .bsuuid is random every time image is run even with same machine-id
hope that random .bsuuid every time doesnt break stuff

efroemling:
Hmm; ideally it would be best if we could get a constant .bsuuid for repeat runs, but it's not the end the world if not. It just means that the v1 master server will get a new account made on it every time someone runs the game which is a bit wasteful.

Loup-Garou911XD:
> Hmm; ideally it would be best if we could get a constant .bsuuid for repeat runs, but it's not the end the world if not. It just means that the v1 master server will get a new account made on it every time someone runs the game which is a bit wasteful.

if we could have a .bsuuid generated in the project somewhere then we can make a volume and access it inside the docker container so it will be consistent with every run

efroemling:
The random .bsuuid happens in core_platform.cc line 161 when needed. Could we add a pcommand that gets run by the docker build after it finishes building the server that just generates one and the same way? It would just be a single line using the python `uuid` module.

efroemling:
I don't really want to make that a part of the standard build because most people should be generating their own .bsuuid at runtime; not picking up a premade one (so that everyone is unique)

Loup-Garou911XD:
> I don't really want to make that a part of the standard build because most people should be generating their own .bsuuid at runtime; not picking up a premade one (so that everyone is unique)

yeah we are not doing that
rn the .bsuuid is randomly generated by the game
what i suggested was generating a .bsuuid outiside of docker image and storing somewhere in project dir and accessing that inside docker by creating a docker volume
this was we can have a consistent bsuuid for a host but different host will have different bsuuid

efroemling:
All I mean is that I think the right way to fix this would be having the Dockerfile run some sort of `tools/pcommand generate_random_bsuuid where/that/should/go/.bsuuid` command itself instead of having that happen as part of `make cmake-build` or whatnot.

Loup-Garou911XD:
> All I mean is that I think the right way to fix this would be having the Dockerfile run some sort of `tools/pcommand generate_random_bsuuid where/that/should/go/.bsuuid` command itself instead of having that happen as part of `make cmake-build` or whatnot.

oh okay now i understand you

Loup-Garou911XD:
we can have a simple script like
```python -c import uuid;print(uuid4()) > .bsuuid && ./ballisticakit_server```
which will generate random bsuuid everytime which btw still happens XD
but ideally i would have a docker volume that has the bsuuid so its consistent and not creating new account on master server every time
and we can also have the config file in this volume for ease of editing

efroemling:
Actually maybe the best thing would be to modify the linux build to use an OS-supplied uuid if there is no .bsuuid file is found; that would fix this case and would require no special code for the docker side of things.

Loup-Garou911XD:
@efroemling i almost forgot
can you hop on discord rq

PR #700, opened 2024-05-18
Very Important Change

### LINKED ISSUE #150: Sticky Bomb Gravitational Pull Bug (Assault)
So when a Spaz goes towards the flag with a sticky bomb stuck to it's head, it goes to its original spawn point but goes to the opposite direction very fast and also more Spazes get interrupted and killed.

vishal332008:
Closes #150


efroemling:
Thanks for this!

PR #706, opened 2024-05-30
Fixing a few UI stuff.

vishal332008:
The mistakes here have been found on different times during different days, so don't ask why it isn't in a single commit. I'd like to end my description here. Thank you.


efroemling:
Thanks for this!

PR #715, opened 2024-07-01
Die Message Fix - Update spaz.py

VinniTR:
bs.DieMessage fix


EraOSBeta (changes requested):
Remove `wasdead` and its usage since this would make that obsolete

efroemling:
Unfortunately I believe this actually breaks things subtly. DieMessage can be used to kill something (allowing for a ragdoll death, fade-out, etc.) but DieMessage(immediate=True) is an alternate form that will kill a thing AND immediately remove it from the game. I feel that the immediate=True form should result in removing something from the game even if its already dead, but with this change that would no longer happen.

efroemling:
Ok looks good now; thanks!

PR #718, opened 2024-07-25
Fixing a bug related to Plugin File Names.

vishal332008:
Fixed a bug where plugins don't run due to them being named same as inbuilt module from standard library of python.
Example : sys.py, os.py, etc.
Please check if this messes with anything else because this getclass function is used in other files too.


efroemling:
This seems like a good check to have; thanks!
However can I request one change?: the getclass() function is supposed to be a generic thing that theoretically could be used with stuff in stdlib modules. Could we move this check up a level so the plugin-specific code makes the check *before* it calls getclass()?

efroemling:
Alternately I'd be fine with having the check be in getclass but controlled by an optional arg like `check_sys_modules_name_clash` which defaults to False.

vishal332008:
> Alternately I'd be fine with having the check be in getclass but controlled by an optional arg like `check_sdlib_modules_name_clash` which defaults to False.

I made some changes, hoping they are correct.

efroemling:
Looks good; thanks!

PR #722, opened 2024-08-25
Revert to ProactorEventLoop on Windows

brostosjoined:
Tested the change and magically there isn't `set_wakeup_fd` errors

### DIFF
--- a/src/assets/ba_data/python/babase/_env.py
+++ b/src/assets/ba_data/python/babase/_env.py
@@ -154,11 +154,11 @@ def on_main_thread_start_app() -> None:
     # error only started showing up once I moved Python init to the main
     # thread; previously the various asyncio bg thread loops were
     # working fine (maybe something caused them to default to selector
-    # in that case?..
-    if sys.platform == 'win32':
-        import asyncio
+    # # in that case?..
+    # if sys.platform == 'win32':
+    #     import asyncio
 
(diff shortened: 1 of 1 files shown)


efroemling:
Can you give me a little background on this (or point me to some if it's out there already)? And which version of windows do you see the set_wakeup_fd errors?

brostosjoined:
So on previous bombsquad version you switced to using `WindowsSelectorEventLoop` due to the error `set_wakeup_fd` only works in the main thread, on test build `1.7.37` i reverted the change to the default windows event loop which is the `ProactorEventLoop`
and found that it works fine without the `set_wakeup_fd` errors. 
I made this pull request to revert the loop to use the default which is the `ProactorEventLoop` so in future we detect what causes the errors maybe debug and fix it.

brostosjoined:
On the mod `we` made `discord_richpresence` to display what your currently doing on the game. I used external libraries to connect discord's api to bombsquad's.The libraries use the defualt event loops for each OS, this causes the game to conflict with the games code hence errors. I then have to manually edit the libraries scripts again inside the mod script making it tedious, modders then sometimes lose intrest.

efroemling:
I went ahead and did this change on my own in the latest build, so I'll go ahead and close this. Thanks!

brostosjoined:
@efroemling would you mind checking your dm

PR #725, opened 2024-09-02
Add official Black formatter to workflow

brostosjoined:
## Steps
Adds a black formarter to the repo's workflow to ensure ci and cd yml  are passing.
I made the changes based on the [wiki](https://github.com/efroemling/ballistica/wiki#:~:text=efrotools.code.black_base_args()).

```python
def black_base_args(projroot: Path) -> list[str]:
    """Build base args for running black Python formatting."""
    from efrotools.pyver import PYVER, get_project_python_executable

    pyver = 'py' + PYVER.replace('.', '')
    if len(pyver) != 5:
        raise RuntimeError('Py version filtering err.')

    return [
        get_project_python_executable(projroot),
        '-m',
        'black',
        '--target-version',
        pyver,
        '--line-length',
        '80',
        '--skip-string-normalization',
    ]
equivalent to 
```

``` yaml
black_args: ". --check --target-version py312 --line-length 80 --skip-string-normalization"
```
Here is the link to black documentation https://black.readthedocs.io/en/stable/integrations/github_actions.html
## Type of Changes

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: rickstaa/action-black@v1
         with:
-          black_args: ". --check"
+          black_args: ". --check --target-version py312 --line-length 80 --skip-string-normalization"
           
   make_linux_x86_64_gui_debug_build:
     runs-on: ubuntu-latest
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: rickstaa/action-black@v1
         with:
-          black_args: ". --check --target-version py312 --line-length 80 --skip-string-normalization"
+          black_args: ". --target-version py312 --line-length 80 --skip-string-normalization"
           
   make_linux_x86_64_gui_debug_build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


EraOSBeta (changes requested):
I think this should only be done on CI and not on CD since CI is for running checks (like this) while CD is only for deployment regardless of whether or not checks pass

brostosjoined:
I reverted the change i think its fine now.

brostosjoined:
My stoopid head*ss though CD was for server builds and CI was for gui.

EraOSBeta (approved):

efroemling:
Thanks for putting this together, but unless I'm misunderstanding something here I don't think this is the right way to address things.
If CI is failing due to badly formatted code then we should be guiding the user how to fix it before checking it in. (`make preflight` will run formatting as well as making sure everything is good to go). If we're formatting code while we run tests then we're just hiding problems; not fixing them since the repo will still contain badly formatted code.
Do you think we need better docs or error messages telling people about `make preflight` or `make format` or whatnot?

brostosjoined:
We can add `--check --verbose` but i think `make preflight`  is superior so ill close this.

brostosjoined:
I was making this so contributors (specifically windows users) who want to make a simple contribution wouldnt have to go through all the configurations and all the other processes of setting up a viable environment for bombsquad

PR #734, opened 2024-09-24
Here we go again, fixing a bug.... made by me in attempt to fix another bug

vishal332008:
1. Fixing a bug I made in last commit.
2. Added a way to change position, height and width for scoreboard.
3. Stopped Stress-Tests from running when we are already present in a game.
4. Fixed a bug which was not showing the winner of FFA when the player left right before the winner display.

(Test it out once just for confirmation).


src/assets/ba_data/python/bauiv1/_hooks.py
```diff
@@ -125,7 +125,13 @@ def root_ui_level_icon_press() -> None:
 
 def root_ui_menu_button_press() -> None:
     from bauiv1._appsubsystem import UIV1AppSubsystem
+    from babase import app
+
+    if app.classic is None:
+        logging.exception('Classic not present.')
+        return
 
+    app.classic.resume()
```
efroemling, line 134: Is there a bug that this is addressing? In general I'm trying to untangle classic gameplay stuff from ui stuff so if resume needs to be called somewhere it should probably go under classic somewhere.

vishal332008: So the bug is, when you are in game and you open the in game menu, it pauses the game, but instead of pressing resume to resume the game, try to press the  top right button which opens the in game button again, it will close the in game menu but doesn't resume the game.

So in the previous fix, I added the __del__ function to resume whenever the in game menu is deleted, but it turns out it created another problem. So now pressing the top right button when the in game menu is open closes the in game menu and resumes it, but if the in game menu is open and you press the settings button, it quickly opens and closes the settings window because the __del__ function runs and the game resumes, so the window is closed.

I thought instead of doing something to the in game menu window, I could just go to the button and resume it when it's pressed so no problems will occur. If you got any other ideas to fix it, it will be nice.

vishal332008:
I did everything I could and I do not know how to fix this part.

vishal332008:
Why does only my PR get affected everytime you make a commit :/

src/assets/ba_data/python/bauiv1/_hooks.py
efroemling: Thanks for the explanation. So I removed the resume() call from the __del__() call and added one at the callback for the menu button in classic._appmode._root_ui_menu_press(). So I *think* it should cover all cases now. Let me know if you find any case still broken...

vishal332008:
@efroemling , I removed my part of the code, so I guess that's it? or is there anything else?
(and check discord, nothing important at all but just saying)

efroemling:
Thanks for this. I'll tidy up the mypy/typing stuff.
One general request for future stuff is to try and add some code comments, especially in places like the multiteamvictory changes where a bit of complexity is being added and its not immediately clear what's going on. I'll take care of it now though.

src/assets/ba_data/python/bascenev1lib/activity/multiteamvictory.py
```diff
@@ -372,39 +377,40 @@ def on_begin(self) -> None:
         v_offs = 0.0
         tdelay += len(player_entries) * 8 * t_incr
         for _score, name, prec in player_entries:
-            tdelay -= 4 * t_incr
```
efroemling, line 375: Question: what is this `if prec.player.in_game` check for? Does it break something to show players who have left? This seems kinda opposite of the code added to show the winner even after they've left.

PR #744, opened 2024-11-09
Fix Multi-Grab Issue

VinniTR:
fix multi-grab issue

current behavior
https://github.com/user-attachments/assets/d8367a45-78bb-42b2-bc8f-1f8a3d94a6e1

new behavior
https://github.com/user-attachments/assets/c1c05208-8ffb-4e66-92cf-f119fa46e0bc

## Type of Changes

### DIFF
--- a/src/assets/ba_data/python/bascenev1lib/actor/spaz.py
+++ b/src/assets/ba_data/python/bascenev1lib/actor/spaz.py
@@ -709,9 +709,10 @@ class Spaz(bs.Actor):
         assert not self.expired
 
         if isinstance(msg, bs.PickedUpMessage):
-            if self.node:
-                self.node.handlemessage('hurt_sound')
-                self.node.handlemessage('picked_up')
+            if not self.node or self.node.hold_node:
+                return None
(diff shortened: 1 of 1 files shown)


EraOSBeta (approved):

VinniTR:
@efroemling after multiple tests I came to the conclusion that this could not be completely solved, when you quickly grab a bomb near a character it is marked as grabbed but we did not grab it, I thought I had easily solved it by changing the AND for OR but Another problem with the tnt and the flag appeared, this definitely has to be fixed in the c ++ layer.

EraOSBeta:
The real issue is how a grab operates: you would normally want the grab to simply affect the first thing it collides with and be over, but since Eric wanted to prioritize flags over other objects in a grab, we have to go through every collision the grab makes and the problems caused by this prioritization method don't have any clean fixes, even in the native layer. Another funny byproduct of this method is that `bs.DroppedMessage`s are dropped for objects that are replaced while `bs.PickedUpMessages`s aren't.

Either way the prioritization of flags in grabs is something necessary but we should discuss what other methods can be used for doing it

PR #757, opened 2024-12-08
Small fix

vishal332008:
It should select the line as meta line only when it starts with `# ba_meta ...` and not if `# ba_meta ` is included in it.
Breaks some plugins such as api updaters and stuff which uses (this is an example -> ) `line.replace('# ba_meta require api 7', '# ba_meta require api 8')` .
As you can see the example, it has `# ba_meta ` in it but it's not a meta line, so yeah. This update should probably fix it.


efroemling:
Thanks for this! I'll accept it but I think I'll modify it slightly after so it does `l.strip().startswith()` instead of just `l.startswith()`. That will let meta tags continue to work on indented sections.

PR #763, opened 2024-12-12
Spawning time fix

vishal332008:
In football, if you grab flag and leave it in a place other than goal area and spawn point of flag, it'll count down from 10 to 1 and then spawn back in it's spawn point, but there is a delay between it disappearing and before it spawns back. Removed that delay so now it should just disappear and spawn back instantaneously.


vishal332008:
@efroemling check PRs, it's getting out of hand. From just 9 to 17 now.

vishal332008:
@efroemling come online in discord, need to have a little talk, just a small talk, but very important.

efroemling:
Thanks!

PR #774, opened 2025-01-03
Update CONTRIBUTORS.md

komasio71:


CONTRIBUTORS.md
```diff
@@ -65,4 +65,7 @@
 ### Loup Garou
 - Added sphinx documentation generation
 - Added docker build system
-- Various CI/CD improvements
\ No newline at end of file
+- Various CI/CD improvements
+
+### komasio71
+- Contributor
```
EraOSBeta, line 71: too vague, be more specific so it becomes funny

EraOSBeta (changes requested):

CONTRIBUTORS.md
komasio71: > too vague, be more specific so it becomes funny

check the new one

PR #781, opened 2025-01-16
Removing Timer for Chests for Gold Pass Players

vishal332008:
So Eric, is this good? I have went through the code a bit and thought a lot and this is the best way.

Before:
image

After:
image


efroemling:
Good catch, and thanks for this!

PR #814, opened 2025-02-22
Customizable time for Freezing

vishal332008:
Allows customizable time for freezing players instead of just 5 seconds designed only for ice bombs.


efroemling:
Thanks!

PR #816, opened 2025-03-09
Fixed Exploit empty device spec packet

HypeStratex:
Fixing Exploit of sending empty "device" spec packets, was making servers crash with -11 error code.


efroemling:
Thanks for this!
I'm applying this fix but in a slightly more general way - instead of specifically checking for and ignoring empty objects I'm simply ignoring everything that is *not* a string; otherwise I think they could trigger the same crash by sending an int or various other things.

PR #826, opened 2025-04-27
Small change

vishal332008:


src/assets/ba_data/python/bauiv1lib/profile/edit.py
```diff
@@ -84,12 +84,8 @@ def __init__(
             root_widget=bui.containerwidget(
                 size=(width, height + top_extra),
                 scale=self._base_scale,
-                stack_offset=(
-                    (0, 0) if uiscale is bui.UIScale.SMALL else (0, 0)
-                ),
-                toolbar_visibility=(
-                    None if uiscale is bui.UIScale.SMALL else 'menu_full'
```
EraOSBeta, line 91: Why remove this?

vishal332008: Um because it's an editing window? Means you shouldn't move to some other window while you're editing, should either save it or cancel it, only then go to some other window

EraOSBeta: Do you always have a single window open on your computer? going to toolbar menus is the equivalent of minimizing your current window, you can always come back to it when you're done.

EraOSBeta: Literally just said that in the previous comment.

vishal332008: Yeah yeah ik 😭 GitHub email delay bro sorry 😭

efroemling: I do think with the new toolbar stuff it does in some cases make sense to hide all the options to switch away to other things and instead force either 'apply' or 'cancel' on the thing currently being done. The game playlist edit UIs are an example of this. So I think it seems reasonable here too.

efroemling:
Thanks for this!

PR #830, opened 2025-05-14
Party Window - copying messages without usernames

SoK05:
## Description
I always found it weird that copying messages also copies usernames. 
Now it doesn't. Enjoy actually copying only the contents of the message :)

## Type of Changes

QoL change.

## Quick Demo

https://github.com/user-attachments/assets/f7b1c35b-7138-4d0c-a061-c6eb009fbbe4


SoK05:
> Very useful for spam as you have showed 😭

You could still spam with usernames. Also there's already spam prevention on parties so whatever xD

PR #704, opened 2024-05-24
init rpc stuff

EraOSBeta:
## My Tasks

- [x] Add Pypresence and setup Discord RPC.
- [ ] Add Discord tab to the gather menu when a Discord client is detected.
- [ ] Do more testing.
- [ ] Remove OVERCLOCKED stuff.
- [ ] Ensure `make preflight` completes successfully.

## Eric Stuff

- [ ] Make an application on your Discord account and use its `Client ID` instead of OVERCLOCKED's
- [ ] upload an asset called `logo` to the above Discord application
- [ ] Make sure I didn't fuck up security-wise when making this
- [ ] Add and use a bunch of `Lstr`s
- [ ] properly license [qwertyquerty/pypresence](https://github.com/qwertyquerty/pypresence) (placed it in /src/assets/ba_data/python/baplus)
- [ ] Add a CHANGELOG entry. 

## Description

This PR adds Discord Rich Presence to BombSquad.

Features:

- Showing info about the player's play session on the player's Discord profile.
- Support inviting and joining other players through Discord.

## Type of Changes

|     | Type                   |
|-----|------------------------|
| ✓   | :sparkles: New feature |

PR #831, opened 2025-05-15
Discord social sdk

Loup-Garou911XD:
## Description
Adding discord social sdk for features like rich presence and more( but mostly focused on rich presence for now)
This exposes 3 functions to python layer under _babase
- discord_start
- discord_is_ready
- discord_richpresence

**Note:** Anyone who wants to try this needs to unzip the [sdk](https://www.jiocloud.com/l/?u=wUkLYuVHKFwHFJMEdi_1e2gwDfOcnMCCKJKcDxTgLDwJcOb6LrTIw-f1PoEOUXtKMHW) in lib/ directory

By default the discord sdk is not included in the build
For including it:
-  use `BA_ENABLE_DISCORD=1` env var
-  change `#define BA_ENABLE_DISCORD 0` in src/ballistica/shared/buildconfig/buildconfig_cmake.h to `#define BA_ENABLE_DISCORD 1`

once running u can run `import _babase;_babase.discord_start();while(!_babase.discord_is_ready()):pass;_babase.discord_richpresence(large_image_key="globe")`

### for @efroemling 
securely store accessToken


uwu-user:
# Well, hello there 🙂
> I know I'm a little annoyed about this, and `I know` I've already reply in something similar to this **[ on [#704](https://github.com/efroemling/ballistica/pull/704) ]** , but it's a little important about what bombsquad and discord have to do with official stuff, not about mods like: `(RichPresence.py)`

---

## - I have some questions

- **[1]** Why is there no **[OFF/NO]** button inside the game settings for this feature "if I don't want it", isn't there supposed to be a button in the Advanced menu in the settings?

> [!CAUTION]
> > like, Sometimes I want to be `"offline"` I dont want to get spam from friends [Because I mod on some servers] Or I don't want to reply to someone and keep myself offline, but when this feature is there, I missed it, I will show that I'm playing bombsquad

- **[2]** Why do you use one `APPLICATION_ID` for this?, If you want to make it public for all bombsquad players `[0.5M+ Daily Player]` you should not use any ID, plus using `APPLICATION_ID` more than 1000 times per minute causes the ID to be banned immediately *[ Source: [discord/rate-limits](https://discord.com/developers/docs/topics/rate-limits) ]*

> [!IMPORTANT]  
> > Just imagine if you were to make an `APPLICATION_ID` every hour because the players are more than the `APPLICATION_ID` could handle, this would be disgusting, or if every player have to make an `APPLICATION_ID` from the discord developers applications for this, this would be a waste of time + This will be difficult for people who don't know anything about discord`

- **[3]** who own's this ID? 
`[src/ballistica/base/discord/discord.h]` `[ line 17 ]`
   ```java 
   static​ ​const​ ​uint64_t​ APPLICATION_ID = ​1373228222002626610​;
   ```

- **[4]** Why is `accessstoken` used if it is available? 
> [!WARNING]
> > it is considered to use the token as a selfbot, which is contrary to Discord's Terms of Service` **[ Source: [discord/self-bot](https://support.discord.com/hc/vi/community/posts/360072076632-discord-selfbots) ]** 

---

## - What is the correct way to do this? To make RichPresence appear in discord officially `(legally)`
- **[1]** Connect Ballistica with your Discord account through the official website of Ballistica via the official OAuth2 service
FromBallistica
> [!IMPORTANT]  
> > It doesn't currently exist, `I know`, but it likes to be like this when there is a way to log in through a discord account or after adding the `Ballistica` authorized accounts method with discord

> [!TIP]
> > Let the work for `eric` because it should do this on the Ballistica site and not through py/java codes**

- **[2]** After adding BombSquad OAuth2, you can find it in Settings from Authorized Apps
AuthorizedApps
> [!IMPORTANT]  
> > It is possible to be `stopped` or `deleted` from the account when you do not want this, it is easy and in a simple way for most players who have no experience about discord...

- **[3]** It can also be stopped without deleting Authorized Apps through a single button inside the game

OnAdvanced

---

> [!IMPORTANT]  
> **know you've put a lot of hard work into this and I appreciate it, but it's just the wrong way and it's not official.**

> [!NOTE]  
> It is better to remind `Eric` to add a login to Discord as a way to create an Ballistica account, as well as a way to link any player's account to a Discord account and not add a random set of codes like this, remember: not all `Bombsquad` players use Discord :>

Loup-Garou911XD:
First i would like to point to the fact that this is still a work in progress hence the draft, which dismisses most of your question
- 1: Its still a work in progress, there will be a on off button in the python and the c++ layer to completely eliminate the sdk as a dependency if someone wants to simplify build process
- 2: Application Id is for Oauth which authorizes our discord application to do some tasks on user's account like setting rich presence, this id is not a secret. After Oauth, we get a id which is client secret which we are currently storing in text file https://github.com/Loup-Garou911XD/ballistica/blob/discordSDK/src%2Fballistica%2Fbase%2Fdiscord%2Fdiscord.cc#L101 because its a work in progress. As for the rate limits, those are for individual clients and we are not making those requests from a single bot 
image
because we exchange application id with a client secret(which is not the standard account token and has limited scope) which is used to make those requests and would be different for each user.I would still confirm this now that you have mention it but yes rate limits should not be a problem
- 3: I own this id, again cause work in progress
- 4: refer point 2

EraOSBeta:
@uwu-user
- [here](https://discord.com/developers/docs/social-sdk/index.html) Discord presents backend-less usage of the SDK as an official option, which means it is not against Discord's ToS no matter what you think.
- (repeat of what Loup said) [here](https://discord.com/developers/docs/discord-social-sdk/getting-started/using-c++) (at step 5) you can see example snippets for implementing the SDK in-game, which uses the application ID. I would like to think that there's no way you can call this the "wrong way" of doing things when it's Discord's own example of how to do it right.

Also if #704 is anything to go by, you will simply repeat your arguments without listening to what others are saying and without doing any additional research when told to do so; so before you make any further comments, know that they will be removed immediately if that were to happen here or if they're unrelated to the PR. In short don't waste your and our time if you aren't going to meaningfully contribute from now on.

uwu-user:
> @Loup-Garou911XD
> First i would like to point to the fact that this is still a work in progress hence the draft, which dismisses most of your question
> 
> * 1: Its still a work in progress, there will be a on off button in the python and the c++ layer to completely eliminate the sdk as a dependency if someone wants to simplify build process
> * 2: Application Id is for Oauth which authorizes our discord application to do some tasks on user's account like setting rich presence, this id is not a secret. After Oauth, we get a id which is client secret which we are currently storing in text file https://github.com/Loup-Garou911XD/ballistica/blob/discordSDK/src%2Fballistica%2Fbase%2Fdiscord%2Fdiscord.cc#L101 because its a work in progress. As for the rate limits, those are for individual clients and we are not making those requests from a single bot
>   image
>   because we exchange application id with a client secret(which is not the standard account token and has limited scope) which is used to make those requests and would be different for each user.I would still confirm this now that you have mention it but yes rate limits should not be a problem
> * 3: I own this id, again cause work in progress
> * 4: refer point 2
> 
> And most of the code here is from discord social sdk's [getting started guide](https://discord.com/developers/docs/discord-social-sdk/getting-started/using-c++) so it is very much the correct and legal way to do it and yeah i havent figured out how to safely store the client secret yet so dont comment on that And even if the client secret is compromised, it wont be much issue because of the limited scope of it

---

> `work in progress` huh? okay then 🤔
- First of all, I know that the `Application_id` is not secret, because I use it in several apps, But it should not be included in the c++/java file, but it should be in a protected file `(.env file. Or something like that)` for safety `(and it Easy to do it, "work in progress" so it fine)`
- Secondly, you should use `Discord OAuth2` I said that it is the wrong way in that you are using `Application_id` not as an app, but as a source of images again

## - If you are asking about the way to create OAuth2, it is simple
1. open [discord developers applications](https://discord.com/developers/applications)
2. click on `"CREATE AN APPLICATION"` and chose a name.
3. at `Installation` page
   > Select `Installation Contexts` on `User Install` and also `Install link` on `Discord Provided Link` and save it.
4. at `OAuth2` page 
   > [ 1 ] - add `https://ballistica.net` on Redirects
   > [ 2 ] - at `OAuth2 URL Generator` mark `identify`, `rpe.activities.write`, `activities.write`, `activities.read` as ✅️
   > [ 3 ] - save it 
   > [ 4 ] - Select `https://ballistica.net` on `SELECT REDIRECT URL`
   > [ 5 ] - Copy the OAuth2 link and Done.

- It should be like this
discordOAuth2

---

The only thing I want is not to put the Discord accounts of most players at risk of banned because it was a 'work in progress' or something like it `(In Discord you do not get a warning and your account is disabled directly If you don't follow the rules, your know that)`
Also this should be working on pc, mac, android and not just on one platform, the reason I say it's the wrong way is because you use the 'application_id` as an image source and `terminal shell` as a way to connect to a discord account this is an old method that works only for PCs,macOS **(I think you already know that)**

> [!TIP] 
> If you plan to do this, also add OAuth2 in `ballistica.net` for some future things, it's a good thing, although a lot of people don't need rpc on BombSquad anymore because it's available `(Available on pc, mac, liunx in several safe ways, in addition to the Kizzy app in Android, Just as a note, but I don't want to talk about them)`

EraOSBeta:
@uwu-user please read Discord's social SDK documentation before continuing to engage.

uwu-user:
> @uwu-user please read Discord's social SDK documentation before continuing to engage.

Don't worry, I've already read it, even though you didn't read my message and deleted it right away on 1s, I'm just thinking about safety first of all!
I know you're mad at me because of what happened earlier, it okay for my but not for your I know 
Also, don't delete my message when I ask for someting, that's not nice

EraOSBeta:
I did read it, and I deleted it because they were concerns that you would know are not actually a problem if you had read the documentation
- This is EXACTLY the way Discord's own SDK documentation says to use the application ID, within the C++ source code as per the the page we linked above, there is no threat to "safety" here
- The asset storage for the application on Discord is meant to be used to load images for backend-less rich presence; that is it's main point for existing in the first place, and once again this is the intended way in Discord's own documentation [here](https://discord.com/developers/docs/discord-social-sdk/development-guides/setting-rich-presence)

So in short I don't believe you read the documentation, in which case you are not welcome to discuss this PR until you do.

uwu-user:
> I did read it, and I deleted it because you were asking for things that you would know are not actually a problem if you had read the documentation
> - This is EXACTLY the way Discord's own SDK documentation says to use the application ID, within the C++ source code as per the the page we linked above, there is no threat to "safety" here
> - The asset storage for the application on Discord is meant to be used to load images for backend-less rich presence; that is it's main point for existing in the first place, and once again this is the intended way in Discord's own documentation [here](https://discord.com/developers/docs/discord-social-sdk/development-guides/setting-rich-presence)
> 
> So in short I don't believe you read the documentation, in which case you are not welcome to discuss this PR until you do.

Well, as you like, I won't talk about this anymore, as you want, 
"But I was just talking about how to make OAuth2", btw how can you link a discord account to this without OAuth2? it makes no sense!!
You must have `OAuth2` linked with your account in order to use `APPLICATION_ID`

on this way: 


Also, I won't talk anymore about this, thank you for your anger and delete the message, bye <3.

Loup-Garou911XD:
@uwu-user 
We already have Oauth2 and its handled by the social sdk itself when we call the authorize method
 
image

brostosjoined:
😂

efroemling:
Thaaaaaaaaanks!

PR #837, opened 2025-06-13
Discord stuff in python layer

Loup-Garou911XD:
## Steps

## Type of Changes


Loup-Garou911XD:
@efroemling the checks are failing 
they pass on when i run make preflight with `BA_APP_RUN_ENABLE_BUILDS=1`
i think it might be because the dummymodules are being generated from upstream:head because that env var is not set

src/assets/ba_data/python/babase/__init__.py
```diff
@@ -200,8 +201,10 @@
 from babase._workspace import WorkspaceSubsystem
 
 _babase.app = app = App()
+_babase.discord = discord = Discord()
```
efroemling, line 204: I'd rather that 'app' be the only object instance directly under babase, so can we store this as `babase.app.discord`? So you'd instantiate it in _app.py instead of here (alongside the various other subsystems).

Loup-Garou911XD:
@efroemling i made the changes u requested
and the tests pass locally

efroemling:
Thanks!!!

PR #840, opened 2025-06-21
generate dummy module docs

Loup-Garou911XD:
## Steps

## Description
we generate docs for dummy modules now
it should be helpful for modders

## Type of Changes

### DIFF
--- a/src/assets/sphinx/template/index.rst_t
+++ b/src/assets/sphinx/template/index.rst_t
@@ -16,6 +16,7 @@ Ballistica Developer's Guide
    commonmodules
    runtimemodules
    toolsmodules
+   dummymodules
 
 
 Indices
--- a/tools/batools/docs.py
+++ b/tools/batools/docs.py
@@ -355,6 +355,26 @@ def generate_sphinx_docs() -> None:
         check=True,
     )
 
+    _printstatus('Generating dummymodules...')
+    subprocess.run(
+        sphinx_apidoc_cmd
+        + [
+            '--doc-project',
+            'Dummy Modules',
+            '--tocfile',
+            'dummymodules',
+            module_first_arg,
+            '--maxdepth',
+            module_list_max_depth,
+            '-f',
+            '-P',
+            str(dummy_modules_filtered_dir),
+        ]
+        + excludes_common,
+        env=environ,
+        check=True,
+    )
+
     # raise RuntimeError('SO FAR SO GOOD')
 
     _printstatus('Running sphinx-build...')


Loup-Garou911XD:
`ballistica/.cache/sphinx/babase._mgen.rst: WARNING: document isn't included in any toctree [toc.not_included]`
getting this warning

Loup-Garou911XD:
how do i include that rst in toctree?

efroemling:
`babase._mgen` is a private module (hence the underscore) so it shouldn't be in docs. Anything from private modules that the user can interact with should be exposed in a public module somewhere. So like how `babase/__init__.py` pulls in most public stuff from `_babase`. All that exposed stuff then shows up in docs.

Is this causing a problem somewhere? I'm assuming all private modules show that same warning?

efroemling:
Looking at this more in general, *everything* currently in dummy-modules is a private module, so all docs for this stuff should be exposed by exposing particular things through public modules as mentioned above (babase for _babase, bascenev1 for _bascenev1, etc). Please holler if you see any shortcomings with that approach. 

*HOWEVER* I am currently working on autogenerating modules for asset-packages and that stuff probably *will* go into dummy-modules (which may or may not be renamed) and those things probably *will* be public modules, so this is still a relevant PR. Perhaps as a test you can explicitly throw a non-private module in there (foo.py, etc) and see what we need to do to get that stuff showing up in the 'Runtime' modules list.

Loup-Garou911XD:
> Looking at this more in general, *everything* currently in dummy-modules is a private module, so all docs for this stuff should be exposed by exposing particular things through public modules as mentioned above (babase for _babase, bascenev1 for _bascenev1, etc). Please holler if you see any shortcomings with that approach. 

Yes I understand those are private modules but many plugins use the internal modules
And having docs for them is helpful

Loup-Garou911XD:
> *HOWEVER* I am currently working on autogenerating modules for asset-packages and that stuff probably *will* go into dummy-modules (which may or may not be renamed) and those things probably *will* be public modules, so this is still a relevant PR. Perhaps as a test you can explicitly throw a non-private module in there (foo.py, etc) and see what we need to do to get that stuff showing up in the 'Runtime' modules list.

So those will be generated from cpp layer?
adding it to the toctree would be difficult

efroemling:
> So those will be generated from cpp layer? adding it to the toctree would be difficult

They will be generated as part of cloud asset builds, so not too different than existing stuff. We'll just wind up with something like `bastdassets.py` in the dummy-modules dir which allows type-safe access to everything in that asset-pack, etc.

Loup-Garou911XD:
> > So those will be generated from cpp layer? adding it to the toctree would be difficult
> 
> They will be generated as part of cloud asset builds, so not too different than existing stuff. We'll just wind up with something like `bastdassets.py` in the dummy-modules dir which allows type-safe access to everything in that asset-pack, etc.
> 

If it helps this case, u can link external docs in sphinx
So maybe if you want to build and host them from your servers, you can just add a link to it in the conf.py just like it links to the base python stuff

efroemling:
> Yes I understand those are private modules but many plugins use the internal modules And having docs for them is helpful

Understandable, but I still think it is a good bit of social engineering to not make docs for internal stuff easily available. If there is a lot of demand for some internal thing, we should polish that thing up and make it public.

My general vision is that *most* modders should have all the functionality they need in their particular feature-set packages (`bascenev1` for gameplay mods, `bauiv1` for ui mods, etc.). Then `babase` exists as a back-up junk drawer of generally-stable-ish stuff if they need it. And finally there's private stuff, but using that should be discouraged as much as possible.

I know I'm often guilty of changing public apis, but I'm trying to move towards a world where public apis stay mostly stable while private apis are free to change. So I think its important to have a clear distinction.

Loup-Garou911XD:
> > Yes I understand those are private modules but many plugins use the internal modules And having docs for them is helpful
> 
> Understandable, but I still think it is a good bit of social engineering to not make docs for internal stuff easily available. If there is a lot of demand for some internal thing, we should polish that thing up and make it public.
> 
> My general vision is that *most* modders should have all the functionality they need in their particular feature-set packages (`bascenev1` for gameplay mods, `bauiv1` for ui mods, etc.). Then `babase` exists as a back-up junk drawer of generally-stable-ish stuff if they need it. And finally there's private stuff, but using that should be discouraged as much as possible.
> 
> I know I'm often guilty of changing public apis, but I'm trying to move towards a world where public apis stay mostly stable while private apis are free to change. So I think its important to have a clear distinction.

Understood, i will close this pr
As for the asset pack docs
I am not entirely sure how that would play out
Ig we could add it to the toctree somehow
Might be useful to look into using the sphinx api instead of cli tools for a cleaner setup
Also might be helpful to check how the jinja templates work

PR #844, opened 2025-07-16
added zstd to apt install in Dockerfile

Loup-Garou911XD:
Fixes the nightly build failing cause of missing zstd in docker


efroemling:
Thanks!

PR #847, opened 2025-08-02
Curse explode message update

vinnytherabbit:


src/assets/ba_data/python/bascenev1lib/actor/bomb.py
```diff
@@ -769,13 +773,17 @@ def __init__(
...
-        # etc. Inheriting owner/source-node from things that set us off
-        # should be all we need I think...
+        # UPDATE (August 2025):
+        # If triggered by another source (e.g., a punch or another bomb),
+        # they may inherit the hit type and subtype of that source,
+        # but only if the incoming subtype matches the bomb's own type.
+        # This restores part of the old behavior (e.g., the "mine punch" trick)
+        # in a safer and more controlled way, preventing unrelated type propagation.
+        # Now, land-mines and impact bombs can behave like punches when appropriate,
+        # without breaking interactions with objects they are meant to destroy.
```
EraOSBeta: I was clearly wrong in assuming that I didn't need to explain this :/
Restoring this logical flaw does 2 things:
1. Brings back functionality that was and still is unintended
2. Changes a mechanism that is meant to be used publicly mid-api version, needing any mods that rely on the old behavior to make two different versions for the same api version

issue 2 could be mitigated; but issue 1 will stand whatever you do: It's unintended behavior; it could be a cool mod but it doesn't have a place in vanilla.

EraOSBeta (changes requested):

src/assets/ba_data/python/bascenev1lib/actor/bomb.py
vinnytherabbit: oops, wrong place.

DinoWattz:
Why are you doing this?

Why doesn't the original code just use `CurseExplodeMessage`?

The heck?

vinnytherabbit:
> Why are you doing this?
> 
> Why doesn't the original code just use `CurseExplodeMessage`?
> 
> The heck?

Well, I made it because I wanted to make it easier to make an explodey with a custom explosion for my modpack, and I think other modpackers would like it.

PR #849, opened 2025-08-02
Bringing back impact/mine-punch old behavior

vinnytherabbit:
I know efro finds it strange, but this is one of the funniest, most bizarre and epic things about Bombsquad. I like the sound of the hit, it makes the attack more epic and not like a normal, boring explosion.

### DIFF
--- a/src/assets/ba_data/python/bascenev1lib/actor/bomb.py
+++ b/src/assets/ba_data/python/bascenev1lib/actor/bomb.py
@@ -1095,7 +1095,9 @@ class Bomb(bs.Actor):
         factory.activate_sound.play(0.5, position=self.node.position)
 
     def _handle_hit(self, msg: bs.HitMessage) -> None:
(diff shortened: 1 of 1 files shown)


EraOSBeta:
Is it funny? Definitely; but it completely messes up the logic (for example the explosion of punched landmines won't blow up other bombs), so I'll close it. You could however turn this into a plugin in the future.

vinnytherabbit:
> Is it funny? Definitely; but it completely messes up the logic (for example the explosion of punched landmines won't blow up other bombs), so I'll close it. You could however turn this into a plugin in the future.

Wait, what? I swore the mines hit were exploding other bombs, it's the msg.radius that prevents it from spreading,  You should try this change to see if this really happened if I suddenly saw something wrong

EraOSBeta:
Oh wait I was wrong; it DOES blow up other bombs, but as far as I can see it won't despawn powerups. Again my point stands: this would be introducing a logical bug just for the sake of "It looked cool"; I shouldn't need to explain why that is a bad idea.

PR #853, opened 2025-08-23
caching python venv for workflow

Loup-Garou911XD:
This pr adds python venv caching to the linux builds
Only linux build because other platform builds dont run as often
here is the comparison in time reduction before and after this change
Notice the time for **Setup project environment**
Screenshot From 2025-08-23 23-34-33
Screenshot From 2025-08-23 23-34-22


Loup-Garou911XD:
@efroemling can u merge this pls
I need the main branch for flatpak workflow

PR #868, opened 2025-10-25
Fix spawn order handling for solo mode

FluffyPal0:
[BUG] In solo elim mode, when a player leaves and their actor alive. As the spawn order is already removed from on_player_leave (because its called first from the Activity) And it'll handle DieMessage of the dying player if its alive and try to remove it while its already removed. Causing ValueError and spaz not dying in the GameActivity.

https://github.com/efroemling/ballistica/blob/main/src/assets/ba_data/python/bascenev1lib/game/elimination.py#L524-L526


efroemling:
Thanks for this!

PR #878, opened 2025-12-09
not erroring out for missing audio device

Loup-Garou911XD:
Idk if there was a reason to crash out for missing audio device
but this fixes an issue i was having


Loup-Garou911XD:
@efroemling if u are free, can you hop on discord

PR #879, opened 2025-12-13
Roundabout - tweaked flag positions

SoK05:
Since the PR with fixing weak gameplay on Happy Thoughts was merged recently, I've decided to do a similar fix for CTF on Roundabout.

Due to flags being too close to the edges, the entire gameplay revolves around jumping to the enemy base and throwing the flag to the other side, scoring a point immediately.

I (and a lot other people) find this strategy weak, and makes players not use the entirety of the map.

This is a simple fix to a big problem.

Before:
Screenshot_2025-12-13_10-53-43

After:
Screenshot_2025-12-13_10-50-04


TheMikirog:
It's not a full fix, since you can have the second player grab the flag on the other side if you jump, but it's a valiant effort and probably the best we can do without changing the map geometry. I approve!

SoK05:
> It's not a full fix, since you can have the second player grab the flag on the other side if you jump, but it's a valiant effort and probably the best we can do without changing the map geometry. I approve!

Yee, because backwards compatibility is important, I couldn't do tweaks to the map itself.. :/

And true, other player can still grab the flag from the other side, but tbh it requires a bit og coordination and teamplay, so it's not as big of an issue as throwing the flag and immediately capping.


But yeah, making gameplay better piece by piece!

I hope we can tweak gloves and shields soon enough too 😁

ZackerDC:
I really like this change because it makes scoring easily more difficult. However, only separating the flags along the X axis still allows that small space to be used to gain momentum and throw the flag. It would be like hiding the usual shortcut a bit more.

https://github.com/user-attachments/assets/6385162f-7cd3-41f8-a258-623a931fc218

On my side, I did some tests by placing the flags in the center of the platform, so it’s no longer possible to take as much advantage of the space to throw the flag that far (it’s still possible, but more complicated).

If you’re looking for a solution that makes it almost impossible to score by just throwing the flag, you can place it in the lower corner closest to the ladder.
With this, the flags are separated, making it almost impossible to score by throwing alone. However, some kind of shortcut to score would appear (a bit more complicated), by making the flag travel across the map to score from below (although this could be fixed by adding collision boxes so it doesn’t happen).

https://github.com/user-attachments/assets/7eaf1a54-b8d7-4c6b-858d-2fa201351e1e

And also, the closest thing to a shortcut would be scoring almost at the end of the ramp, but this already relies more on using the character’s movement than simply throwing it and scoring in 2 seconds.

https://github.com/user-attachments/assets/83ba436b-494b-48c8-9f9b-5c9f6e9886f0

In any case, separating the flags really helps people stop using that annoying shortcut.

SoK05:
> I really like this change because it makes scoring easily more difficult. However, only separating the flags along the X axis still allows that small space to be used to gain momentum and throw the flag. It would be like hiding the usual shortcut a bit more.
> km_20251214_720p_60f_20251214_132620.mp4
> 
> On my side, I did some tests by placing the flags in the center of the platform, so it’s no longer possible to take as much advantage of the space to throw the flag that far (it’s still possible, but more complicated).
> 
> If you’re looking for a solution that makes it almost impossible to score by just throwing the flag, you can place it in the lower corner closest to the ladder. With this, the flags are separated, making it almost impossible to score by throwing alone. However, some kind of shortcut to score would appear (a bit more complicated), by making the flag travel across the map to score from below (although this could be fixed by adding collision boxes so it doesn’t happen).
> km_20251214-1_720p_60f_20251214_133621.mp4
> 
> And also, the closest thing to a shortcut would be scoring almost at the end of the ramp, but this already relies more on using the character’s movement than simply throwing it and scoring in 2 seconds.
> km_20251214-2_720p_60f_20251214_134128.mp4
> 
> In any case, separating the flags really helps people stop using that annoying shortcut.

The reason I decided to keep the flags at the same X axis is making it less obvious which path the attackers will take with the flag. Will they take a ramp, or throw the flag down? Your solution makes throwing the flag impossible, but at the same time it makes the map a bit less interesting.

Yea I see my solution isn't exactly perfect, but I want to see how it looks in the heat of the battle. Doing this big throw will be almost impossible if just one person blocks your way. 

We should still defo look how my change does so we can possibly tweak this in the future, though.

efroemling:
Applied this change to the source maya file, which came through as 9e850a5aace0d5e3d2f77eb80991ce0ccc89f7d0

Thanks!

PR #883, opened 2025-12-25
ci/cd cmake builds

Loup-Garou911XD:
Added `using-cmake` input to the "Setup BombSquad Environment" action to get us setup with required packages and  sccache for building and caching cmake stuff

It does generate alot of cache entries but we should never run into trouble with managing it manually 
If it's ever needed to manage it manually then gh cli tool is quite useful there


efroemling:
Thanks!

PR #886, opened 2025-12-31
Update mainmenu.py

Erfan-0:
Update 2025 to 2026 (The most valuable commit of the year xD)

### DIFF
--- a/src/assets/ba_data/python/bauiv1lib/mainmenu.py
+++ b/src/assets/ba_data/python/bauiv1lib/mainmenu.py
@@ -254,7 +254,7 @@ class MainMenuWindow(bui.MainWindow):
             text=(
                 f'{app.env.engine_version}'
                 f' build {app.env.engine_build_number}.'
-                f' Copyright 2011-2025 Eric Froemling.'
+                f' Copyright 2011-2026 Eric Froemling.'
             ),
             h_align='center',
             v_align='center',


Loup-Garou911XD:
Should we not use
https://efroemling.github.io/ballistica/efrotools.html#efrotools.project.get_non_public_legal_notice instead of hardcodeing it?

HeyErfan:
It definitely seems better, but is there a reason why no one has done it so far?

efroemling:
Thanks!

Loup-Garou911XD:
> Should we not use
> https://efroemling.github.io/ballistica/efrotools.html#efrotools.project.get_non_public_legal_notice instead of hardcodeing it?

@efroemling

efroemling:
> > Should we not use
> > https://efroemling.github.io/ballistica/efrotools.html#efrotools.project.get_non_public_legal_notice instead of hardcodeing it?
> 
> @efroemling

Yes that could be done. Though it would need to be wired up to happen in project-update or something (those tools modules aren't available at runtime) - and its questionable as to whether the time to wire that up would outweight the time it takes to update the hard-coded value once per year 🤔.

Loup-Garou911XD:
> > > Should we not use
> > > https://efroemling.github.io/ballistica/efrotools.html#efrotools.project.get_non_public_legal_notice instead of hardcodeing it?
> > 
> > @efroemling
> 
> Yes that could be done. Though it would need to be wired up to happen in project-update or something (those tools modules aren't available at runtime) - and its questionable as to whether the time to wire that up would outweight the time it takes to update the hard-coded value once per year 🤔.

Updating it once per year is certainly easier then

PR #898, opened 2026-01-26
 Fix docs and added docs badge to README and deployment

Loup-Garou911XD:
Screenshot From 2026-01-26 00-22-31


Loup-Garou911XD:
@efroemling this one too

PR #899, opened 2026-01-26
More cleanup for flathub stuff

Loup-Garou911XD:
fixed some stuff and moved some func from pcommand.py and pcommand2.py to pcommand3.py
now they all are <1k lines


Loup-Garou911XD:
@efroemling merge pls

Issue #882, opened 2025-12-25
Missing audio device causes crash


Loup-Garou911XD:
### Description
Missing audio device causes crash
```
0.482 INFO ba.app: BallisticaKit 1.7.59 build 22677 starting...
ALSA lib confmisc.c:855:(parse_card) cannot find card '0'
ALSA lib conf.c:5204:(_snd_config_evaluate) function snd_func_card_inum returned error: No such file or directory
ALSA lib confmisc.c:422:(snd_func_concat) error evaluating strings
ALSA lib conf.c:5204:(_snd_config_evaluate) function snd_func_concat returned error: No such file or directory
ALSA lib confmisc.c:1342:(snd_func_refer) error evaluating name
ALSA lib conf.c:5204:(_snd_config_evaluate) function snd_func_refer returned error: No such file or directory
ALSA lib conf.c:5727:(snd_config_expand) Evaluate error: No such file or directory
ALSA lib pcm.c:2721:(snd_pcm_open_noupdate) Unknown PCM default
FATAL ERROR: No audio devices found. Do you have speakers/headphones/etc. connected?
----------------------- BALLISTICA-NATIVE-STACK-TRACE-BEGIN --------------------
./run(_ZN10ballistica4core12CorePlatform19GetNativeStackTraceEv+0x65) [0x55fad2058cb5]
./run(_ZN10ballistica18FatalErrorHandling16ReportFatalErrorERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEb+0x1c4) [0x55fad2189924]
./run(_ZN10ballistica18FatalErrorHandling12DoFatalErrorERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE+0xf) [0x55fad218ae3f]
./run(_ZN10ballistica4base11AudioServer6Start_Ev+0x17ca) [0x55fad1f3294a]
./run(_ZN10ballistica8Runnable15RunAndLogErrorsEv+0x2c) [0x55fad219bf6c]
./run(_ZN10ballistica9EventLoop20RunPendingRunnables_Ev+0x75) [0x55fad217e8b5]
./run(_ZN10ballistica9EventLoop4Run_Eb+0x260) [0x55fad2185f90]
./run(_ZN10ballistica9EventLoop17ThreadMainAudioP_EPv+0x5e) [0x55fad21868fe]
[...]

PR #902, opened 2026-02-01
add null audio device

Loup-Garou911XD:
This adds a check and ignores AL commands when no audio device is found and we are using null device(null device means no device found and is not actually an audio sink)

tested only on linux with `ALSOFT_DRIVERS=disabled make cmake`

Screenshot From 2026-02-01 19-41-02

the error in the screenshot is unrelated and caused in gather window


Loup-Garou911XD:
closes #882

Loup-Garou911XD:
Github runners facing outage, rerun when they are back
https://www.githubstatus.com/
image

efroemling:
Thanks for doing this. Looks good to me. My one note is that there are three very similar log messages in your screenshot. I think a single one would look cleaner if there's not a good reason to have the others. Also maybe we should show a screen-message so people with no audio know something is wonky?.. but both of those changes could be in a followup PR so I'll pull this one for now.

Loup-Garou911XD:
> Thanks for doing this. Looks good to me. My one note is that there are three very similar log messages in your screenshot. I think a single one would look cleaner if there's not a good reason to have the others. Also maybe we should show a screen-message so people with no audio know something is wonky?.. but both of those changes could be in a followup PR so I'll pull this one for now.

i fixed the three log messages iirc, that screenshot is old

PR #904, opened 2026-02-06
Update README.md to use relative links for badges

Loup-Garou911XD:
use relative links for status badges in readme so that it displays the fork's status instead of upstream


efroemling:
Thanks for this!

PR #905, opened 2026-02-06
switch to python's native repl

Loup-Garou911XD:
This allows for things like autocompletion, command history, multiline statements and line editing(going left/right)

only tested this on linux

Also if we use autocompletion then gc gets alot of `inspect._signature_fromstr.<locals>.RewriteSymbolics` stuff and complains on exit

```

babase.do_once(                                
>>> babase.d
babase.displaytime(   babase.displaytimer(  babase.do_once(       
>>> babase.d8.476 ba.app: Shutting down...
9.038 ba.gc: Explicit gc pass handled 1650 objects from last 8.5s in 0.011s (total: 1650).
Objects by type:
  tuple: 550
  function: 440
  cell: 330
  dict: 220
  type: 110 (inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.RewriteSymbolics, inspect._signature_fromstr.<locals>.Rewr...

```
https://github.com/user-attachments/assets/62f201ae-73a7-415d-9c11-60b7de5684a3

Screenshot From 2026-02-07 04-38-00

### DIFF
--- a/tools/batools/appmodule.py
+++ b/tools/batools/appmodule.py
@@ -278,7 +278,9 @@ def _generate_app_mode_selection(
     return out
 
 
-def _generate_default_imports(out: str, projroot: str, feature_sets: dict[str, FeatureSet], info: str) -> str:
+def _generate_default_imports(
+    out: str, projroot: str, feature_sets: dict[str, FeatureSet], info: str
+) -> str:
     """Generate default imports code for the REPL."""
 
     # Generate default imports in run_default_imports function in _app.py
@@ -299,7 +301,7 @@ def _generate_default_imports(out: str, projroot: str, feature_sets: dict[str, F
             'must be a dict with string keys and string or None values, '
             'or not present'
         )
-    
+
     # Filter imports to only include those from available feature sets
     available_imports = {}
     for module_name, alias in default_imports.items():
@@ -336,7 +338,9 @@ def _generate_default_imports(out: str, projroot: str, feature_sets: dict[str, F
     return out
 
 
-def _generate_repl_help(out: str, projroot: str, feature_sets: dict[str, FeatureSet], info: str) -> str:
+def _generate_repl_help(
+    out: str, projroot: str, feature_sets: dict[str, FeatureSet], info: str
+) -> str:
     """Generate repl_help() function to display available imports and info."""
 
     # Generate repl_help() function to display available imports and info.


efroemling:
Thanks for putting this togather! Awesome to see the live autocomplete vid.

A few initial notes/thoughts:
- From what I understand, readline is not available on all platforms (for example windows) and apparently newer Python repls are migrating away from it to pyrepl. Is that something we can use here instead of readline?
- babase can't depend on bombsquad-specific things (bascenev1, bauiv1, etc) so we can't hard-code imports to them there. I can understand wanting to auto-import things but we'll need that to live in some bombsquad-specific place or just be a user-configurable thing. Either way we should split that off into a separate topic I think.
- On a related note, whenever executing hard-coded Python code from C++ we should make a function for it in a place like babase/_hooks.py and then grab a reference to the function in C++ land and run that. (see src/meta/babasemeta/pyembed/binding_base.py). That way type-checking and linting runs on that Python code and so CI will tell us when we're doing something we shouldn't be doing like importing bauiv1 from babase code.
- For something like this which might need some testing or have weird corner cases I'd recommend we make it configurable so we can have it be an opt-in thing at first and eventually switch it to the default. Do you think you could add an entry for it in base/support/app_config.cc (something like 'Enable Native REPL') and key off of that, with it falling back to the old behavior when disabled?

Loup-Garou911XD:
Thanks for the feedback 
1. I could definitely try to add pyrepl instead of readline
2. Could we have a function like default_imports in babase._hooks for the imports, because that would be bombsquad specific and not in the engine itself

And yeah I will try to make all the requested changes, soon™

Loup-Garou911XD:
@efroemling also to do about the gc warns

efroemling:
The base feature-set doesn't rely on any others (bascenev1, etc) so we can't hard-code any usage of those things anywhere in it (including babase._hooks). My internal CI tests enforce this by running checks on all feature-sets with non-dependencies stripped out. So since babase depends on nothing but core, it has to be able to pass tests with bascenev1 and all that gone. I should move those tests to github actions at some point.

What we *can* do is something like how project-update auto-generates parts of babase/_app.py based on the project-config. So we could have something like "default_imports" in config/projectconfig.json and those would get written into a 'run_default_imports()' method in babase/_app.py or something like that, which would allow us to give a default set of imports for bombsquad or any other spinoff project. Though I also think it could make sense to make that user-configurable. A few years from now we might not want to waste cycles importing bascenev1 if we're then on bascenev4 or whatnot. Also I want to be really careful to avoid situations where something *relies* on one of those auto-imports happening, which is a big part of why I'm not auto-importing anything currently.

As far as the GC warns: I wouldn't worry about that. We can either try to break those particular reference cycles or just silence the warnings in those cases. The warnings are mainly to minimize the amount of GC happening during gameplay as that can cause hitches or memory bloating, but I think if someone's mucking around in the console then those things are more acceptable.

efroemling:
On a related note, I like how blender's python console gives a little blurb about auto-imported modules, convenience variables, etc.

image

Maybe we could have something similar, or a blurb about how to add your own auto-imports or something.

Loup-Garou911XD:
> On a related note, I like how blender's python console gives a little blurb about auto-imported modules, convenience variables, etc.
> 
> image
> 
> Maybe we could have something similar, or a blurb about how to add your own auto-imports or something.

Nice idea

Also I like u went from a maya boy to a blender boy

src/ballistica/base/support/stdio_console.cc
```diff
@@ -32,79 +39,126 @@ void StdioConsole::StartInMainThread_() {
...
+  // _pyrepl is still very new and is not well documented.
+
+  Python::ScopedInterpreterLock gil;
+
+  // Enable readline for full REPL features:
+  // - Command history (up/down arrows)
+  // - Line editing (backspace, left/right)
+  // - Tab autocompletion
+  const char* readline_import =
+      "import readline\n"
```
Loup-Garou911XD: @efroemling is this fine being here or do i need to move it to _hooks.py ??

efroemling: Its a pretty simple bit of Python code so it wouldn't be terrible to do it inline this way, but I'd suggest making a _hooks.py entry for it. That way you can see how that stuff works. Plus, if there winds up being more setup lines necessary like readline.parse_and_bind() then they're all properly type-checked.

If you do want to keep it as an inline PyRun_SimpleSting() though that's fine. The one thing I'd suggest in that case though is to log an error or just FatalError() if the PyRun_SimpleString() fails (returns -1). Its never good to let substantial things fail silently. PyRun_SimpleString() will print a stack trace to stderr on exceptions but stderr might not be easily viewable on all platforms.

Loup-Garou911XD:
@efroemling now the game doesnt shut down properly
idk what the issue is
~its only happening after the merge~

Edit: i fixed it

Loup-Garou911XD:
okay so if its using the fall back to the old repl
and we exit the game
it prints `>>> 3.797 ba.app: Shutting down...`
but doesnt exit
but if we press enter after the shutdown prompt it exits

Edit: i fixed it

Loup-Garou911XD:
@efroemling this can be merged if you like it

Loup-Garou911XD:
maybe ill make a follow up pr with that blurbs out the auto imported stuff

config/projectconfig.json
```diff
@@ -22,6 +22,20 @@
...
+    "bauiv1",
+    "bauiv1lib"
+  ],
+  "default_imports_aliases": {
+    "babase": "ba",
+    "bascenev1": "bs",
+    "bascenev1lib": "bs_lib",
+    "bauiv1": "bui",
+    "bauiv1lib": "bui_lib"
+  },
```
efroemling: It seems like for these sort of handy utility imports we will very often be providing aliases. Could we simply have default_imports be the dict with aliases? (with None specifying no alias).

src/assets/ba_data/python/babase/_app.py
```diff
@@ -1204,6 +1204,45 @@ async def _fade_and_shutdown_audio(self) -> None:
...
+        _mod = importlib.import_module('bascenev1lib')
+        main_dict['bascenev1lib'] = _mod
+        main_dict['bs_lib'] = _mod
+        _mod = importlib.import_module('bauiv1')
+        main_dict['bauiv1'] = _mod
+        main_dict['bui'] = _mod
+        _mod = importlib.import_module('bauiv1lib')
+        main_dict['bauiv1lib'] = _mod
+        main_dict['bui_lib'] = _mod
+
```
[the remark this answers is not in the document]
Loup-Garou911XD: Iirc, I was getting some error with the standard import foo as f
Don't remember what the error was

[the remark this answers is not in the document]
Loup-Garou911XD: it doesnt run in `__main__` in fact
`assert __name__=='__main__'` raises error

Loup-Garou911XD:
@efroemling a little help with the spinoff error
also added a function that prints module names as you requested
lmk if you want any changes in that
and this pr can be merged once the error is fixed and the gc warns are taken care of
image

efroemling:
> @efroemling a little help with the spinoff error also added a function that prints module names

module name printing looks good - my only request is to change 'Available modules:' to 'Auto-imported modules:' (available modules sounds ambiguous to me - like it is ones I could import if I wanted to or something)

As far as spinoff errors: spinoff is basically creating a new project by stripping out feature-sets of this one. That test is stripping out everything except `base` and making sure things still build. Look for code that auto-generates other parts of _app.py and I think you'll find examples of code being generated based on what feature-sets are available. In this case you'll want to only apply an auto-include if the associated feature-set is still intact. Holler if you get stuck on this part and I can give more specifics.

Loup-Garou911XD:
> > @efroemling a little help with the spinoff error also added a function that prints module names
> 
> module name printing looks good - my only request is to change 'Available modules:' to 'Auto-imported modules:' (available modules sounds ambiguous to me - like it is ones I could import if I wanted to or something)
> 
> As far as spinoff errors: spinoff is basically creating a new project by stripping out feature-sets of this one. That test is stripping out everything except `base` and making sure things still build. Look for code that auto-generates other parts of _app.py and I think you'll find examples of code being generated based on what feature-sets are available. In this case you'll want to only apply an auto-include if the associated feature-set is still intact. Holler if you get stuck on this part and I can give more specifics.

done and done
image

efroemling:
Awesome! Is this mergeable now or are you still working?

Loup-Garou911XD:
> Awesome! Is this mergeable now or are you still working?

yeah should be if ci passes

Loup-Garou911XD:
> Awesome! Is this mergeable now or are you still working?

also dont forget the gc warnings
i didnt take care of those

Loup-Garou911XD:
ahh 💩 
i forgot to run `make preflight`

Loup-Garou911XD:
okay should be good now

Loup-Garou911XD:
✅ green light for merge

efroemling:
Thanks for this!

The one nitpicky change I will probably make is switching the setting name from "Use native python REPL" to ""Use Native Python REPL" to match our weird capitalized setting name convention.

PR #909, opened 2026-02-12
get some assets needed in cmake linux env

Loup-Garou911XD:
flathub builds fail because of these


Loup-Garou911XD:
@efroemling we would need a new tag push to get the new linux build env in release
is that possible?

PR #911, opened 2026-02-14
Volume slider values are now better matching user expectations

TheMikirog:
Previously, volume values were more a suggestion rather than an estimate, since these applied linearly to volume. Unfortunately, this meant high values closer to 100% barely changed the perceived volume, while low values closer to 0% were impossible to fine tune. If that previously annoyed anybody, now it's gone! **50% is now half as loud as 100%**. This change will be most noticeable on big speakers and in the future if volume adjustment settings get slider widgets instead of plus and minus buttons. 🔊

In terms of technical details, linear volume scale gets raised to the power of 3, which matches closer to audio sliders used in DAWs, plugins, and synths. There's not one consensus on what is the best algorithm, but this felt better within the context of the game and user interface than true logarithmic or power of 4. I aimed for good enough than the best possible. 📓

I did not touch the volume scales used in sound functions on the Python layer, since those have already been set in stone and applying my change on a per-sound source basis risks throwing off the sound mixing. All that was changed here were set volume functions in the audio server.

Tested on Windows. Though my change appears platform agnostic, this also needs testing on Mac, Linux and Android.


efroemling:
Thanks for this!

PR #923, opened 2026-03-25
Prevent punches from being allowed before grabbing opponents

TheMikirog:
**"Punch grab spam"** (or "infinite" or "exploit" as I'll be calling it for clarity) has been the most **destructive and polarizing technique for the game's health**. It involves abusing an exploit where players can instantly kill one another by landing a single punch and grab, keeping them in grasp as their health dwindles with continuous punches they can do nothing about. This keeps going until death, or forever if target can be stunned, but is invincible. 

After many debates with the community, @efroemling confirmed that he wants this exploit fixed, so here is my attempt that solves the problem it at its core.

## How to do it in game

1. Build up speed for the initial stunning hit.
2. Press the Grab button, then **QUICKLY** press the Punch button.
3. Land both the grab and the punch on the opponent. If done correctly, the character should be able to damage the victim as they're holding it.
4. Build up speed by jumping and/or doing spins. Keep doing this from that point on.
5. Rhythmically repeat these inputs: Bomb/Grab/Punch. Bomb releases the character, while the other buttons allow for a regrab and dealing damage (and more stun). If done in a particular rhythm, you can maximize damage while keeping the victim stunned and unable to act.
6. Repeat ad infinitum.

## How the exploit works
Infinite abuses a quirk related to grabbing mechanics. When the grab button is pressed, a grab animation starts. The player then spawns a grab hitbox after 4 game ticks have passed, which is where the actual grab happens. However, **there was no check in the game that prevented you from punching before then**. This oversight allows the player to grab the opponent, and then punch them shortly after (hence the common name for the exploit).

## Why this is bad
Most melee fights started revolving around this trick, which suffocated the rest of the strategy and techniques that lead to a more satisfying, fun, and long lasting combat loop, for both players and spectators. Due to the exploit's importance in the game's meta, it became necessary to learn in order to compete. This leads to a boring game that is also unintuitive and unfriendly for beginners. It's worth noting that most people don't use this exploit, but those who do, kill the enjoyment for everyone involved, since it generates great frustration due to the lack of counter-play and big punishment for simple mistakes. Infinite also kills gamemodes that are relying on players being invincible but can be stunned. This necessitates gamemode specific fixes, which can't fix this exploit outright and often require sacrificing game feel. Removing this exploit from the game will open it up for more interesting fights, and a more intuitive game that's easier to learn, yet still provides ample opportunity for execution based tricks (more on this later).

## The fix
This fix has been implemented in the internal layer, where the bug lies. When a punch is about to commence, the game first checks if a grab has been initiated beforehand. If so, don't punch. However, once that 4 tick grace period for the grab's hitbox spawn ends, punches are possible again.

## Compatibility and Implementation
Fix has been implemented under `behavior_version` number 2 for Spaz node. Fix can be reverted by setting `behavior_version` to a lower value. There's also the attribute `pickup_before_hitbox`, which allows the Python code to recognize when a grab was started, but before the grab actually happens.

We tested the fix by having an old build client join the fixed build host, and there were no issues in terms of desyncs and other network incompatibilities. However polishing the mechanic in terms of visuals and code openness necessitated creating new SpazNode attributes, which might warrant a new protocol.

If there's something that could be improved here in terms of code is a cleaner way to merge internal and Python code together. In order to maintain compatibility, I put the fix under a `behavior_version` check. However, I used `pickup_before_hitbox` to disallow the punch swish sound from being played while the glitchy punch is happening. The fix is a bit hacky because of this, and Python is dealing with certain audio visual stuff that I think it should not be dealing with. Most of the grab and punch logic happens in the internal layer. Bridging this gap while maintaining modding compatibility will be difficult.

## Gameplay consequences
For the regular player that never used this exploit, **they won't notice any changes while playing**.
However, this exploit is a fundamental fix to the game's core mechanics, and as such **it impacts other physics techniques that players used to rely on**. Players are understandably worried and/or curious about if those techniques are still possible, and if there are any possible workarounds. Good news is, most tricks are still possible, and are either harder to execute and/or less potent. I was more comfortable leaving these variants in due to their difficulty and impact on the game.

- **Infinite** has been nerfed in two ways. Now it requires reversed inputs (Punch, then Grab), which means its harder to do and requires stricter rhythm and timing. In our experimentation, it's also impossible to loop. Trying to loop it results in less damage, less stun, and requires the victim to already be stunned in order to land one iteration of the loop. One time someone got a two-hit combo with this technique that almost resulted in death, and looked cool to boot. In our testing, we could not pull off a version of the trick that resulted from full health to death, and kept the victim inactionable throughout. In this version of the trick, **the victim has agency**, since they get unstunned and can retaliate while being held. This was the missing piece required that made the technique better to deal with and more understandable to the player.
- **Bomb jumps** (punching bomb to give it momentum while regrabbing it in mid-air to fly) also requires reversed inputs and stricter timing. For the average player, a single bomb jump will give less height. However with good enough execution, you can match the height and distance from before the fix.
- **Bomb drops** (spawning a bomb already on the ground) remain unchanged, since that trick does not rely on the grab exploit to function.

Most of these techniques being harder to do might require playing on a keyboard or fighting stick due to the precision and timing required, which puts gamepad and touch screens at a disadvantage. However, since the infinite is not necessary for competition, it should put all input methods closer together in terms of competitive viability.

## Conclusion
This exploit being fixed marks the end of an era for competitive BombSquad. It will take some time before the community fully adjusts, but I firmly believe that if the game wants to grow in the future with its 2.0 launch, it can't afford to scare away new players that would've otherwise really enjoyed the game, or even played it competitively. I've had the pleasure to test this fix extensively with tons of lovely people and it's refreshing to show people how deep the game can be, and what would be sacrificed by keeping the exploit unfixed.

**Huge thanks** to @GummyBoiYT1, @SoK05, @NeoArcadiax, and a bunch of other people that tried out this fix while it was in development. Thanks to you this fix could be done as accurately and tastefully as possible, while avoiding side effects.


efroemling:
Thanks!

IvanPragma:
I have to push back on this one.

This "bug" has been in the game for about ~10 years. At this point it's not an exploit, it's a feature the competitive scene was built on. An entire layer of tech depends on the grab→punch timing window: not just bomb jumps, but throws/bombs-shooting and a bunch of other movement and combat tricks. The fix doesn't delete them, sure, but it forces everyone to relearn muscle memory built over a decade, for no real gain.

The claim that "only a handful use it and everyone hates it" doesn't match reality. In my circle the entire pro community is used to it and relies on it. It's not some fringe griefing tool, it's core meta.

And here's the actual problem: **this fix doesn't fix anything**. The "kill a player in your hands" exploit still works, you said it yourself, just swap grab→punch for punch→grab. On touchscreen maybe not (or really hard), but on gamepad/keyboard it's easy. The fact that you couldn't loop it isn't proof it's gone — that's a skill issue. I've attached a clip with `behavior_version=2` on both spazes where the exploit is clearly still usable. So all this really does is make the same thing harder and input-order-dependent, while breaking everything else.

My proposal: **leave it as is**. It's a feature, not a bug. At a serious competitive level it gives no real advantage. Forcing the reverse input only makes the game worse, near-impossible to pull off on phones, and a forced relearn for every PC/gamepad player who put in the years.

https://github.com/user-attachments/assets/e8586b09-bf5e-4444-b722-094248be7ccd

TheMikirog:
@IvanPragma After months of the community playing around with this fix and debating, it's clear that it's not perfect.

For context, this fix was optional for a while now, hidden behind the newest network protocol for servers. The average person was not able to test these changes thoroughly, though people that have been involved with me trying to find the fix reported its shortcomings after this fix went live.

The biggest issue with it right now is that it impacted too many of the core mechanics that weren't as contentious (bomb jumps and bomb drops being the main sticking points). People did not have issues with these, just the infinite. It impacts a lot of interactions in a negative manner, but the rest are either situational or could be fixed elsewhere (like Assault requiring a CTF flag reset for players to score).

Currently I'm investigating different ways of addressing the infinite. It's clear that the grab hitbox coming before the punch hitbox creates emergent gameplay in a lot of places, some of which are desirable. Ideally we want to get rid of the infinite while keeping the execution of other tricks the same if possible. It's very plausible the final iteration of these changes will be in the Python layer, though I'm still researching this in my free time. This fix will definitely not become the default until all of those concerns have been addressed.

I do not agree with your "competitive people use it and its not a problem there" argument though, as we've been dealing with this one ever since the exploit first became popular, and that never held water to me.

> The claim that "only a handful use it and everyone hates it" doesn't match reality. In my circle the entire pro community is used to it and relies on it. It's not some fringe griefing tool, it's core meta.

In my original post I argued this trick is harmful to players of all skill levels, not just average players (I outline this in the "Why this is bad" section). While this statement you reference is an exaggeration, it does indeed get closer to reality, because **pro competitive community is not the majority of players**. You can't make general statements by looking at a narrow slice of the community. As a game designer, you have to look at everybody that plays the game. The people that play on servers and play coop make a huge chunk of the playerbase. It was pretty clear that outside of the competitive scene, this exploit was negatively impacting average player enjoyment every time it popped up. 

As for the technique having its place in high level play, **just because something is meta doesn't mean it's good**. If all it takes to get into high level play is to learn this one technique, then **what separates the best of the best from just good players**? This isn't like Smash Bros Melee's wavedashing that genuinely expands the game with more possibilities. Infinite instead takes away from the game at higher levels of play.
- Singular non-grabbing punches lose utility, as there is a more reliable way of securing a kill
- Singular non-punching grabs lose utility, as the risk of being instantly killed yourself is far higher if a single hit can be lethal
- Neutral and footsies matter less, since positioning when using the infinite is far more lenient, and exchanges involving the trick are guaranteed to end in one hit
- The "run for damage but become predictable" balance is completely negated if you have a pocket insta-kill with you at all times that does not require building up speed to pull off
- Huge throws look much more fun to the average player than choking someone with the infinite, despite both of them resulting in a kill and both of them having the victim unable to escape

A game can be more shallow with more mechanics if most mechanics go unused. A game's overall complexity is not an unmitigated good.

"Appeal to tradition" and "appeal to authority" arguments cannot trump a whole bucket list of other arguments that have not yet been defeated. As I outlined this before and will continue to outline it: the main goal is to make the game enjoyable for the **average** player, the majority. If the game is better at the cost of a niche, so be it.

Infinite will be fixed. Currently it's a matter of how, not if.
Still, I appreciate your concerns and rest assured, we'll keep iterating on this!

PR #936, opened 2026-06-12
Update actions

Loup-Garou911XD:
This pr updates action versions to get rid of the deprecated nodejs version 20 warnings
And i also updated every other action to latest version while i was doing that


efroemling:
Thanks!

PR #942, opened 2026-07-11
Adding Rotate option to Image Widget

vishal332008:
and also fixing a comment, spinner widget was named image widget


vishal332008:
@efroemling how is this?

efroemling:
Thanks for this!

PR #946, opened 2026-07-23
Rotate Prop Node 

vishal332008:


Dliwk:
I think currently this gives only visual rotation, which does not affect physics

PR #950, opened 2026-07-29
Allow image node to be placed in_world

Dliwk:
I'm not sure whether I should bump protocol version to 40 or describe changes in 39, so I left `scene_v1.h` as is.

Adds `in_world` attribute to image node, works just as in text node. 

<details>
<summary>screenshot</summary>
image
</details>


efroemling:
Thanks for this! I'll be bumping the protocol version to 40 soon so that'll cover this.

brostosjoined:
Using assets now players can have logos

PR #951, opened 2026-08-02
Punches now deal no damage to other Spaz characters shortly before and after grabbing

TheMikirog:
## Context
This is a response to my earlier pull request #923.
I highly recommend reading that thread for more context and reasoning behind these new changes.
Here is an abridged list of points made previously:

**Punch grab infinite** is an exploit destructive to the game's health, where a player can continuously deal damage and stun to the victim while grabbing, making them unable to escape. My previous pull request aimed to address it.
That fix targeted the game's core grabbing mechanics, which prevented a punch from connecting after the grab button is pressed, but before it actually happens. However this fix was not perfect for two big reasons.

1. Alternative methods for doing this trick that supposedly made it balanced, actually introduced an **uneven playing field across input devices**. The trick was harder to do overall, but easier to do on gamepad and keyboard after button remapping.
2. This fix **impacted other existing techniques** that were not nearly as contentious amongst players, like "bomb jumps" and "bomb punches". If those ever become a problem in the future, it's better to patch them individually, rather than fix them through fundamental changes to system mechanics.

Due to these shortcomings, a new fix was required.

## The New Fix
Old fix where punch hitbox is not created if punch input is pressed before grab finishes **has been removed**.
In its place, a new fix has been implemented.

The game checks for punch eligibility every time a punch connects with the victim.
There are two conditions that must be met in order for the punch to deal damage.
- Time difference between between pressing grab and then punch must be 70ms or later.
- Time difference between punching and releasing the victim from a grab must be 360ms or later.

These values are tailored to timings when these tricks are commonly performed, while making sure just dropping the victim to punch them, without a re-grab, is not impacted. For good measure, I made it so that the maximum stun length Spaz can get when getting hit is lower. This should address frustrating situations where Spaz gets a big hit and is stunned for longer than reasonable, sometimes for random durations.

## Implementation
In the internal layer, I added a `pickup_release_time_ms` attribute, which is a game timestamp describing when something has been released from a grab. That attribute is then read in the Python layer to implement the actual punch prevention fix. Increased stun duration is tied to behavior version 2.

Since the old fix existed for a relatively short amount of time, I was not worried with breaking compatibility with Spaz node's behavior version 2. Some logic from the old fix remains in case it proves useful, such as the new behavior version for the Spaz node, and the `pickup_before_hitbox` attribute. The former should allow for more experimentation in Spaz logic while keeping 1.7 compatible, while the latter might be used for modders if they want to implement logic done when pressing Grab+Punch.

## Conclusion
As with any major change, I encourage people to push this fix to its limits, in order to find situations where it doesn't do its job, or infringes on some crucial aspect of gameplay.

Huge thanks to `tin1ss` on Discord for their contributions when brainstorming ideas for the new fix!


slightlyepic:
IMO infinite punch grab (or at the very least punch grab) is central to many non-combat game mechanics, especially in epic mode and on non-vanilla servers. I don't think this is a good idea. I'd be interested to hear the general opinion on this topic.

SoK05:
> IMO infinite punch grab (or at the very least punch grab) is central to many non-combat game mechanics, especially in epic mode and on non-vanilla servers. I don't think this is a good idea. I'd be interested to hear the general opinion on this topic.

Infinite punch grab is hurtful to the game and makes it flat af. I don't wanna dwell on that anymore because we had a shoitton of discussions on Ballistica Discord. It's tiring. Eric himself said it's existence is not intentional and is okay with removing it. If you say it's central to the game, then sorry to say - skill issue, yadda yadda.

slightlyepic:
I see your point, which is reasonable, but changing mechanics after a decade seems wrong even if said mechanics were not put there intentionally. Although this opinion is probably the result of my very narrow gameplay tastes. You guys and Erik have a broader perspective.

Dliwk:
Can you run:
```bash
git reset --soft HEAD^^^^^^
git restore --staged .
git add <list of files with actual changes>
git commit 
git push --force
```
This will squash all 6 commits into one, but it will also remove these unrelated formatting changes. I'm sure you didn't mean to add those. If something goes wrong, you can use `git reflog` to return to the starting point.

TheMikirog:
> Can you run:
> 
> ```shell
> git reset --soft HEAD^^^^^^
> git restore --staged .
> git add <list of files with actual changes>
> git commit 
> git push --force
> ```
> 
> This will squash all 6 commits into one, but it will also remove these unrelated formatting changes. I'm sure you didn't mean to add those. If something goes wrong, you can use `git reflog` to return to the starting point.

Running preflight is standard protocol, while other commits are separate to keep track of my changes and revert if things go wrong. I meant every commit that I sent with the PR.

src/ballistica/base/graphics/texture/ktx.cc
```diff
@@ -321,7 +321,7 @@ typedef short int16;
 #define PUTBITS(dest, data, size, startpos) \
   dest = ((dest & ~MASK(size, startpos))    \
           | ((data << SHIFT(size, startpos)) & MASK(size, startpos)))
-#define SHIFTHIGH(size, startpos) (((startpos) - 32) - (size) + 1)
+#define SHIFTHIGH(size, startpos) (((startpos)-32) - (size) + 1)
```
Dliwk, line 324: These are definitely not related. If you scroll through all changes, you will see a lot of them.

TheMikirog: Preflight does them, and it does not care if it's relevant to the commit or not.

PR #969, opened 2026-08-31
add instant replay api

Loup-Garou911XD:
added instant replay feature, and also add keyframe to brp so we can start stream from the middle
still needs a cleanup pass, and some rigorous testing
also one side effect might be that we have large .brp files, would have to see how much and is it worth it

https://github.com/user-attachments/assets/4958fcc2-f790-472f-896d-2348287604df

- need to cleanup the skip interface, and maybe make it like tutorial skip


Loup-Garou911XD:
@efroemling check this out

nooboholic-SC:
Instead of this automatic popup to skip
can u add a "Record That" button in pause menu 
which will save last 30 sec of game play

TheMikirog:
I approve of this being something available to modders, especially if we can create a good feature out of it in the future.
Ideas float around already like a play of the game style showcase of the best moment in a session after it ends.

However, I definitely disapprove of it being something that is shown in the middle of gameplay.
Mostly because interrupting the game for everyone for little reason will be confusing for newbies and annoying for vets.
Even if you skip the replay, people stuck watching it can't play the game or contribute to the team.
Mid-gameplay replays definitely work better in turn based games (Worms Armageddon) or games where everyone gets their own screen and the player would otherwise do nothing if the replay was not playing (kill cams in shooters).

EraOSBeta:
I'm in favor of having an API for saving specific parts of the recording to memory for future use, being able to start replays from specific places in the record, and optionally some freedom for messing with the replay scene with the only restriction being that replay objects and after-effects objects lack physically interactions with each other if someone wants to tackle that. I'm against **any** sort of functionality added with this that faces the end-user as of right now, even if it ends up being opt-in. we can have people make concepts using this API as mods and only incorporate something officially if/when we find a good concept with good reception; but adding anything besides the API right now would be too rash in my opinion.

Loup-Garou911XD:
> I'm in favor of having an API for saving specific parts of the recording to memory for future use, being able to start replays from specific places in the record, and optionally some freedom for messing with the replay scene with the only restriction being that replay objects and after-effects objects lack physically interactions with each other if someone wants to tackle that. I'm against **any** sort of functionality added with this that faces the end-user as of right now, even if it ends up being opt-in. we can have people make concepts using this API as mods and only incorporate something officially if/when we find a good concept with good reception; but adding anything besides the API right now would be too rash in my opinion.

Yeah I understand, this was just a proof of concept implementation for the engine.
I haven't given any thought to any functionality facing the end user.
And considering the suggestions from everyone, i will only add the api in this pr and leave out any user facing functionality as future work 👍

